PR #655, opened 2026-05-02
fix(publish): avoid SIGPIPE on enumerate step in publish-all-packages.yml

jeremylongshore:
`jq ... | head -20` triggered SIGPIPE once `head` had read 20 lines, which under `set -o pipefail` aborted the whole enumerate step with exit code 2 even though the listing succeeded. Switched to jq's `.[:20]` slice so there's no pipe to break.

`publish-all-packages.yml` is the manual-dispatch emergency mass-publish workflow (still in use post-#647 as the escape hatch when the incremental publish flow needs a one-shot fallback). Three-line fix, no behavior change beyond not aborting the enumerate.

jeremy made me do it
-claude


jeremylongshore:
Closing — investigated and confirmed this is a duplicate of #544 (commit `742c2a04b`), which shipped the identical SIGPIPE fix. Local branch was a never-PR'd parallel copy of the same work. The fix IS in production via #544.

jeremylongshore.com made me do it
  -claude
intentsolutions.io

PR #737, opened 2026-05-17
catalog: add aomi entry to marketplace.{json,extended.json}

CeciliaZ030:
Closes the prescreen blocker on #727: *"Missing catalog entry for plugins/crypto/aomi in marketplace.extended.json"*.

## What landed

- `marketplace.extended.json` — new aomi entry with verification block (Grade A / 94 from the marketplace-tier validator on `plugins/crypto/aomi/skills/transact/SKILL.md` after today's frontmatter fixes on aomi-labs/skills:main)
- `marketplace.json` — mirror entry, `totalPlugins` bumped to 429
- `scripts/validate-catalog-invariants.py` passes

## Context

The aomi bundle was materialized into `plugins/crypto/aomi/` via `sources.yaml` in #679 (merged 2026-05-12), but the catalog entries weren't added at the time. PR #727 lifted aomi-transact's SKILL.md from grade C → grade A; this PR completes the registration so the prescreen no longer hard-blocks.

The materialized SKILL.md updates from today (frontmatter fixes, ## Resources section, scoped Bash, semver versions) are flowing in via the daily sources.yaml sync from aomi-labs/skills:main.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #818, opened 2026-06-01
catalog: add aomi entry to marketplace.{json,extended.json}

victorchimakanu:
Hi Jeremy — re-filing the catalog entry from the now-closed #737, rebased onto current `main` so there are no conflicts this time. Thanks for the note that a fresh PR was welcome.

**What this does**
- Adds the `aomi` plugin bundle to both `marketplace.json` and `marketplace.extended.json` under `category: crypto`.
- Bumps `totalPlugins` 425 → 426.
- The two skills (`aomi-transact`, `aomi-build`) already merged into your store via the earlier sync; the plugin source is already present at `plugins/crypto/aomi/` (`.source.json` synced from `aomi-labs/skills`). This PR is **catalog-entry-only** — it just makes the bundle discoverable in the marketplace listing.

**On verification:** I left the `verification` block off the extended entry on purpose — happy for your prescreen to assign the grade rather than self-asserting one. If you'd rather I add it back to match the schema, say the word and I'll push it.

Minimal diff: 2 files, +50/-1. JSON validated locally (both files parse, plugin count consistent). Holler if you want it split or tweaked.

### DIFF
--- a/.claude-plugin/marketplace.extended.json
+++ b/.claude-plugin/marketplace.extended.json
@@ -9953,6 +9953,31 @@
         "badge": "gold",
         "lastValidated": "2026-05-20T00:00:00.000Z"
       }
+    },
+    {
+      "name": "aomi",
+      "source": "./plugins/crypto/aomi",
+      "description": "Aomi for AI agents: drive natural-language EVM transactions (chat, fork-simulate, sign) and scaffold new Aomi apps from API specs. Non-custodial, account-abstraction-first. 40+ protocol apps: Polymarket, 1inch, GMX, Hyperliquid, Aave, Uniswap. Bundle ships two skills: aomi-transact and aomi-build.",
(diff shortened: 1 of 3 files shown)


jeremylongshore:
Merged! Thanks for picking this up — the catalog entry was sitting in the gap between the sources.yaml auto-sync (which only does filesystem) and the manual catalog step. Closes one of the 16 stranded entries I'd been tracking.

Pushed a follow-up commit to the branch before merge to align `marketplace.json` with `sync-marketplace.cjs` output (the `totalPlugins` field had drifted on main — your hand-edit was correct in intent, just one number off relative to current main; sync normalized it). README TOC regen picked up the aomi row in the Crypto & Web3 table automatically.

Cheers — appreciate the clean rebase + the deferential approach on the verification block.

- Jeremy Longshore
intentsolutions.io

PR #826, opened 2026-06-03
fix(skill-paths): resolve ${CLAUDE_SKILL_DIR} paths to ./ for GitHub browse view

jeremylongshore:
## Summary

Federico Sapuppo reported via LinkedIn DM (2026-05-28) that SKILL.md files across the repo use `${CLAUDE_SKILL_DIR}/...` path references that don't resolve when browsing the GitHub source. The targeted files DO exist on disk, but a browser-visiting user can't click through or copy a working path. The example he cited was `soc2-audit-helper` but the pattern is widespread.

## The fix

Replace the variable form with relative paths everywhere it's used as a path. Works for both audiences:

- **GitHub browse view:** `./references/foo.md` is human-readable and the path resolves correctly relative to the SKILL.md file viewer.
- **Claude runtime:** when reading a SKILL.md, the working-directory context IS the skill directory, so `./references/foo.md` resolves to `<skill-dir>/references/foo.md` (canonical Anthropic spec).

## Transforms applied

| Before | After | Count |
|---|---|---|
| `${CLAUDE_SKILL_DIR}/../` | `../` | 7 |
| `${CLAUDE_SKILL_DIR}/` | `./` | 1,200 |
| bare `${CLAUDE_SKILL_DIR}` (no trailing `/`) | **preserved** — teaching-concept docs | 9 |

## What's NOT changed

Bare `${CLAUDE_SKILL_DIR}` (no trailing `/`) is preserved in 9 files where the variable is referenced as a teaching concept, not used as a path:

- `plugins/skill-enhancers/skill-creator/**` — docs that explain the path-variable spec to skill authors
- `plugins/saas-packs/claude-pack/skills/clade-reference-architecture/**` — reference architecture docs
- `plugins/crypto/market-price-tracker/skills/tracking-crypto-prices/ARD.md` — guidance about using the variable

These are documentation about Anthropic's path-variable spec itself, not unresolved paths.

## Verification

Before the change, Federico's example file rendered as:

```markdown
See `${CLAUDE_SKILL_DIR}/references/implementation.md` for the six-phase implementation guide.
```

After:

```markdown
See `./references/implementation.md` for the six-phase implementation guide.
```

Browse-view test: viewing `plugins/security/soc2-audit-helper/skills/assisting-with-soc2-audit-preparation/SKILL.md` on GitHub now shows readable relative paths a contributor can copy and trace.

## Test plan

- [ ] View any SKILL.md in `plugins/security/`, `plugins/saas-packs/`, or `plugins/skill-enhancers/` on github.com and confirm `${CLAUDE_SKILL_DIR}/` no longer appears in prose path references.
- [ ] Spot-check that bare `${CLAUDE_SKILL_DIR}` (no trailing slash) still appears in `plugins/skill-enhancers/skill-creator/skills/skill-creator/SKILL.md` where it's documenting the spec.
- [ ] Run any one of the affected skills to confirm Claude's path resolution still works (e.g., `soc2-audit-helper`).

## Scope: 266 files, 1,207 path replacements

Touches every skill pack that adopted the `${CLAUDE_SKILL_DIR}` pattern: security pack (soc2-audit-helper, etc.), saas-packs (40+ vendor packs), ai-ml, skill-enhancers, crypto, testing, productivity, governance. Generated content (`marketplace/src/data/skills-catalog.json`) is not touched because it's regenerated from source SKILL.md files at build time.

- Jeremy Longshore
intentsolutions.io

### DIFF
--- a/plugins/ai-agency/shipwright/package.json
+++ b/plugins/ai-agency/shipwright/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@intentsolutionsio/shipwright",
-  "version": "1.0.1",
+  "version": "1.0.2",
   "description": "Describe your app in plain English — Shipwright builds, tests, and deploys it autonomously via a 9-phase pipeline.",
   "keywords": [
     "app-builder",
--- a/plugins/ai-agency/shipwright/skills/shipwright-pipeline/SKILL.md
+++ b/plugins/ai-agency/shipwright/skills/shipwright-pipeline/SKILL.md
@@ -84,4 +84,4 @@ Shipwright converts a plain-English app description into a fully built, tested, 
 
 - [Shipwright on GitHub](https://github.com/Wynelson94/shipwright)
 - [product-agent on PyPI](https://pypi.org/project/product-agent/)
-- See `${CLAUDE_SKILL_DIR}/references/examples.md` for usage examples.
+- See `./references/examples.md` for usage examples.
--- a/plugins/ai-ml/ai-sdk-agents/package.json
+++ b/plugins/ai-ml/ai-sdk-agents/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@intentsolutionsio/ai-sdk-agents",
-  "version": "1.0.3",
+  "version": "1.0.4",
   "description": "Multi-agent orchestration with AI SDK v5 - handoffs, routing, and coordination for any AI provider (OpenAI, Anthropic, Google)",
   "keywords": [
     "ai-sdk",
--- a/plugins/ai-ml/ai-sdk-agents/skills/orchestrating-multi-agent-systems/SKILL.md
+++ b/plugins/ai-ml/ai-sdk-agents/skills/orchestrating-multi-agent-systems/SKILL.md
@@ -47,7 +47,7 @@ Design and implement multi-agent systems using AI SDK v5 with structured handoff
 9. Implement circuit breakers and timeout guards to prevent workflow deadlocks
 10. Test each agent in isolation, then validate end-to-end handoff chains with representative inputs
 
-See `${CLAUDE_SKILL_DIR}/references/implementation.md` for the detailed implementation guide.
+See `./references/implementation.md` for the detailed implementation guide.
 
 ## Output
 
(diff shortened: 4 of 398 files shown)


jeremylongshore:
Documenting why this one's on hold before it merges — and the shape the fix should take.

Context: this came from Federico Sapuppo flagging (via LinkedIn) that `${CLAUDE_SKILL_DIR}/…` renders as dead text when you browse a SKILL.md on GitHub — you can't click it or copy a working path. That's a real papercut and a good catch.

The snag is that `${CLAUDE_SKILL_DIR}` does **two different jobs**, and they need different treatment:

**1. Doc / reference links** (`${CLAUDE_SKILL_DIR}/references/foo.md`, assets, examples) → these *should* become plain relative paths. That's exactly what the open standard prescribes — AgentSkills.io shows `[the reference guide](references/REFERENCE.md)` and `scripts/extract.py` as bare relative paths from the skill root ([agentskills.io/specification](https://agentskills.io/specification) → "File references"). Converting these makes us *more* spec-compliant.

**2. Script execution** (`python ${CLAUDE_SKILL_DIR}/scripts/foo.py`, `bash …/scripts/foo.sh`) → these need to **keep** the variable. When Claude executes a bundled script the working directory isn't guaranteed to be the skill folder, so `./scripts/foo.py` can resolve to the wrong place and fail. That's the entire reason the variable exists, per the Claude Code docs: *"Use this in bash injection commands to reference scripts or files bundled with the skill, regardless of the current working directory"* ([code.claude.com/docs/en/skills](https://code.claude.com/docs/en/skills) — the `${CLAUDE_SKILL_DIR}` substitution row + the `python3 ${CLAUDE_SKILL_DIR}/scripts/visualize.py` example).

As it stands, this pass converts both: it turned ~295 `python ${CLAUDE_SKILL_DIR}/scripts/…` calls into `python ./scripts/…`, which would break those skills at runtime (e.g. `arbitrage-opportunity-finder`).

**The version that lands clean:** convert the doc/reference links to relative, but leave `${CLAUDE_SKILL_DIR}` anywhere a command actually executes a script (`/scripts/`, `/bin/`, fenced bash blocks). Same browse-ability win, zero runtime breakage. Holding this PR until it's scoped that way; full credit to Federico for surfacing it.

- Jeremy Longshore
intentsolutions.io

PR #823, opened 2026-06-03
docs(databricks-pack): draft cost-leak-hunter CFO output spec for @Gingiris-1031 review

jeremylongshore:
## Summary

Sub-deliverable from the issue [#795](https://github.com/jeremylongshore/claude-code-plugins-plus-skills/issues/795) thread with @Gingiris-1031 (Yipei Wei). Closes the **spec-design half** of bead `claude-0co4`. The skill code work itself stays gated on her sign-off — that's by design.

## Why this is a separate PR (not just a comment)

The spec is a **binding output contract** if Yipei signs off — it becomes what `databricks-cost-leak-hunter` is required to emit. That's substantial enough to live as a versioned doc in the pack's `000-docs/` design-record stream, not as a transient GitHub comment. A reviewer (now or later) needs to be able to read the same thing the skill is built against.

If she redlines or rejects, the doc gets superseded — but the audit trail of what we proposed and what got cut survives.

## What's in the spec

New file: `plugins/saas-packs/databricks-pack/000-docs/014-DR-DRFT-cost-leak-cfo-output-spec.md`

### The structure (AFFiNE-derived)
One sentence what + one proof. For CFO audience the proof is a **dollar number + date range**.

### Per-category output template
```
$X,XXX/month wasted on N <unit>, week of <YYYY-MM-DD>, fix in Y <atomic action>.
```

Four categories, each with format + worked example + evidence table:
1. **All-purpose vs job cluster overspend** ([D07](../blob/main/plugins/saas-packs/databricks-pack/000-docs/002-RL-RSRC-databricks-compute-pain-research.md))
2. **Instance-pool `min_idle` dual-billing trap** ([D09](../blob/main/plugins/saas-packs/databricks-pack/000-docs/002-RL-RSRC-databricks-compute-pain-research.md))
3. **DLT serverless misuse** ([D08, D11](../blob/main/plugins/saas-packs/databricks-pack/000-docs/003-RL-RSRC-databricks-delta-streaming-research.md))
4. **Idle SQL warehouse** ([D11](../blob/main/plugins/saas-packs/databricks-pack/000-docs/002-RL-RSRC-databricks-compute-pain-research.md))

Plus a summary rollup at the top of skill output and explicit rendering order (sorted highest-waste → lowest).

### What's explicitly cut from the CFO output
- Cluster topology diagrams
- DBU efficiency ratios / utilization %
- Spark UI execution plans
- Photon stats, partition counts, bytes-shuffled
[...]

PR #820, opened 2026-06-02
Add skillrot to external plugin sources

mishanefedov:
## Summary

Adds [**skillrot**](https://github.com/mishanefedov/skillrot) to `sources.yaml` so it's mirrored daily into the community marketplace.

skillrot is a linter that audits coding-agent skills for **external-CLI drift**: it reads each skill's bash, extracts the `command subcommand --flags` it uses, and checks them against the installed CLI's own `--help`. When a tool ships a new version and renames a flag or drops a subcommand, the skill keeps emitting the dead command and fails mid-task — skillrot catches that before it bites. Read-only (only ever runs `--help`/`--version`). Also ships a `--cost` context-budget audit measuring the always-on token tax each skill adds per session.

- Repo: https://github.com/mishanefedov/skillrot (MIT)
- Source path: `skills/skillrot` (mirrors `SKILL.md`)
- Works across Claude Code, Codex, opencode, Cursor — any `SKILL.md`-format agent

## Test plan

- [x] `sources.yaml` parses as valid YAML (49 sources)
- [x] Entry follows the existing external-source schema (name/description/repo/source_path/target_path/author/license/category/include/exclude)
- [ ] Maintainer: daily sync mirrors `skills/skillrot/SKILL.md` into `plugins/community/skillrot`

PR #811, opened 2026-05-31
Add google-maps-agent-skills to community plugins

manchungkali:
Adds google-maps-agent-skills to the Community section.

8 skills for Google Maps lead generation: scrape businesses, extract emails, find local leads, analyze competitors, write cold outreach.

Install: `npx skills add gmapsscraper/google-maps-agent-skills/google-maps-scraper`

Repo: https://github.com/gmapsscraper/google-maps-agent-skills

PR #838, opened 2026-06-04
feat(ci): pr-classifier — deterministic file-level PR component detection

jeremylongshore:
## Summary

PR 1 of the 3-PR new-track CI overhaul. Replaces the coarse `awk -F/ '{print $1"/"$2"/"$3}'` depth-3 extractor in `pr-prescreen.yml` that caused doc-only PRs to drag unrelated skills into validator scoring. **PR #823 was the trigger case** — a doc-only edit caused the prescreen to scan all 24 skills in `databricks-pack`, inherit a pre-existing C-grade as the PR's blocker, and `gh pr review --request-changes` against the contributor.

This PR is **foundation only** — adds the classifier + tests, doesn't yet rewire the prescreen workflow. PRs 2 and 3 build on this output to split `validate-plugins.yml` by domain and rewrite the prescreen.

## What ships

| File | Purpose |
|---|---|
| `scripts/pr-classifier/__init__.py` | Package marker |
| `scripts/pr-classifier/rules.py` | Entire deterministic ruleset + catalog/sources diff parsers |
| `scripts/pr-classifier/detect_components.py` | CLI entry point (`--stdin` / `--file` / `--diff-file` / positional) |
| `scripts/pr-classifier/README.md` | Usage + contract + extension instructions |
| `tests/pr-classifier/test_classifier.py` | **46 deterministic tests** |
| `tests/pr-classifier/fixtures/pr-823.{files,diff,expected.json}` | Real-PR snapshot regression for the bug-triggering case |

## Key design points

- **Depth-4 cut for skills.** `plugins/<cat>/<plugin>/skills/<skill>/SKILL.md` surfaces ONLY the named skill. No cascade to sibling skills.
- **Deterministic.** Same input → same output, always. No network, no LLM, no I/O beyond `--file` / `--diff-file` reads.
- **JSON-serializable, sorted lists, stable across runs.** Verified by `test_deterministic_repeat_runs`.
- **15 contribution types** covered: skill, agent, mcp, hook, plugin, catalog, catalog_add, sources, sources_add, ci, frontend, script, test, doc, plus an `unknown` detector that surfaces unrecognized patterns.
- **Self-classifies.** The classifier classifies its own PR correctly — `test_self_classification` pins this.

## Output JSON shape

```json
{
  "contribution_types": ["plugin", "skill"],
  "plugin_paths": ["plugins/saas-packs/databricks-pack"],
  "affected_skills": ["databricks-incident-runbook"],
  "affected_agents": [],
  "affected_mcp": [],
  "affected_hooks": [],
  "catalog_additions": [
    {"name": "aomi", "source": "./plugins/crypto/aomi", "category": "crypto"}
  ],
  "sources_additions": [],
  "file_categories": {"md": 4, "json": 2},
  "touches_workflows": false,
  "touches_frontend": false,
  "touches_scripts": false,
  "touches_tests": false,
  "unknown": false,
  "unmatched": []
}
```

## Bug caught while writing the tests

`marketplace/DESIGN.md` and similar `marketplace/<not-src>/*.md` files were not being classified as docs because the doc rule excluded ALL of `marketplace/`. Fixed in this PR: only `marketplace/src/**` is treated as frontend; the rest of `marketplace/` counts as doc. The test `test_marketplace_root_not_src_doesnt_count_as_frontend` pins the corrected behavior.

## How CI picks this up

Wires into the existing `python-tests` matrix job in `.github/workflows/validate-plugins.yml` automatically — `test_*.py` is auto-discovered. **No new workflow needed for this PR.**

## Test plan

- [ ] `python3 -m pytest tests/pr-classifier/ -v` shows 46 passing tests.
- [ ] Smoke-test the CLI:
  - `python3 scripts/pr-classifier/detect_components.py plugins/security/penetration-tester/skills/auditing-npm-dependencies/SKILL.md --pretty` outputs `affected_skills: ["auditing-npm-dependencies"]`
  - `python3 scripts/pr-classifier/detect_components.py --file tests/pr-classifier/fixtures/pr-823.files --diff-file tests/pr-classifier/fixtures/pr-823.diff --pretty` matches the committed expected.json
- [ ] CI's `python-tests` job runs the 46 new tests + the existing repo tests without regression.

## What's NOT in this PR

- **Workflow split** — `validate-plugins.yml` still monolithic. PR 2 does that, keyed on this classifier's output.
- **Prescreen rewrite** — `pr-prescreen.yml` still uses the old `awk` extractor. PR 3 rewires it to consume this classifier's output and post a single coordinated comment.
- **Public grading page** — PR 3 ships the `tonsofskills.com/grading` page using the same rubric the prescreen grades against.
- **More historical snapshot fixtures** — only #823 is captured here. PRs #811, #818, #819, #820, #821, #822 can land as fixtures in follow-up PRs.

## Why this PR is small + reviewable

The plan called for shipping the classifier independently so PRs 2 and 3 can each be reviewed without cross-cutting churn. After this lands, PR 2 introduces the new workflow files; PR 3 wires the prescreen to consume the classifier. Each PR is independently reversible.

- Jeremy Longshore
intentsolutions.io

### DIFF
--- a/scripts/pr-classifier/README.md
+++ b/scripts/pr-classifier/README.md
@@ -1,0 +1,132 @@
+# pr-classifier — File-level PR component detection
+
+Given a list of files changed in a PR, emit a structured JSON object describing the contribution: which plugins, skills, agents, MCP servers, hooks, workflows, scripts, and docs are affected, plus any catalog or sources additions parsed from the diff.
+
+Deterministic. Same input → same output, always. No network, no LLM, no I/O beyond `--file` / `--diff-file` reads and stdout JSON emission.
+
+## Why this exists
+
+The previous prescreen pipeline used a coarse `awk -F/ '{print $1"/"$2"/"$3}'` extractor at depth 3, which surfaced the plugin directory of any changed file. That caused doc-only PRs touching `plugins/saas-packs/<pack>/000-docs/foo.md` to drag in every skill in the pack for validator scoring — including unrelated skills that already had pre-existing C grades, which were then incorrectly attributed to the new PR as blockers. PR #823 hit this exactly.
+
+The classifier cuts at depth 4 for skills (`plugins/<cat>/<plugin>/skills/<skill>/SKILL.md`) and only surfaces the SPECIFIC skill whose SKILL.md was modified. Pack-level changes still surface as `plugin_paths`, but the prescreen no longer fans out into every skill in the pack by default.
+
+## Usage
+
+### From stdin (typical CI)
+
+```bash
+git diff --name-only origin/main..HEAD | \
+    python3 scripts/pr-classifier/detect_components.py --stdin
+```
+
+### From a file list
+
+```bash
+python3 scripts/pr-classifier/detect_components.py --file /tmp/pr-files.txt
+```
+
+### With diff for catalog/sources detection
+
+```bash
+python3 scripts/pr-classifier/detect_components.py \
+    --stdin --diff-file /tmp/pr-diff.patch
+```
+
+### Positional
+
+```bash
+python3 scripts/pr-classifier/detect_components.py \
+    plugins/security/penetration-tester/skills/auditing-npm-dependencies/SKILL.md \
+    .github/workflows/validate-plugins.yml
+```
+
+### Pretty-print
+
+```bash
+... --pretty
+```
+
+## Output JSON shape
+
+```json
+{
+  "contribution_types": ["plugin", "skill"],
+  "plugin_paths": ["plugins/saas-packs/databricks-pack"],
+  "affected_skills": ["databricks-incident-runbook"],
+  "affected_agents": [],
+  "affected_mcp": [],
+  "affected_hooks": [],
+  "catalog_additions": [
+    { "name": "aomi", "source": "./plugins/crypto/aomi", "category": "crypto" }
+  ],
+  "sources_additions": [],
+  "file_categories": { "md": 4, "json": 2 },
+  "touches_workflows": false,
+  "touches_frontend": false,
+  "touches_scripts": false,
+  "touches_tests": false,
+  "unknown": false,
+  "unmatched": []
+}
+```
+
+All top-level lists are sorted; the dict is JSON-serializable and stable across runs.
+
+## Contribution types
+
+| Type          | Trigger                                                             |
+| ------------- | ------------------------------------------------------------------- |
+| `skill`       | A `plugins/<cat>/<plugin>/skills/<skill>/SKILL.md` file             |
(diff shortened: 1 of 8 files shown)


jeremylongshore:
## Code review applied (commit fc6b41fbe)

Code-reviewer agent flagged **3 CRITICAL + 4 HIGH + 3 NIT** items. All CRITICAL and HIGH addressed:

**CRITICAL — Brace-in-string corrupts catalog parser.** A description like `"Uses {foo {bar} patterns"` (unbalanced braces inside a JSON string value) silently drove the depth counter wrong and DROPPED the entire catalog entry. Real plugin descriptions use template placeholders in curly braces — this is a live risk. Fixed with `_count_unquoted_braces()` helper that walks char-by-char with string-state tracking (handles `\"` escape).

**CRITICAL — Depth-6+ skills silently misclassified.** The hardcoded `path.parts[3] == "skills"` check missed sub-vendored layouts like `plugins/saas-packs/<vendor>/<sub>/skills/<x>/SKILL.md`. No such plugins exist today but it's a forward-looking risk. Replaced with path-tail signature matching (`parts[-3] == "skills"`), same depth-flex applied to agent rule.

**HIGH — `sources_additions` parser fired on non-root `sources.yaml`.** Inconsistent with `classify_files` which correctly restricts to root. Fixed via regex match against the diff-git header.

**HIGH — `file_categories` was non-deterministic.** Counter preserves insertion order based on input file order. `to_json(sort_keys=True)` hid this at serialization but the in-memory dict broke the "deterministic" contract. Now sorted before return.

**HIGH — Self-classification test listed 5 of 9 files.** Expanded to the full set (including README and fixture files) so the test pins what the classifier actually sees on its own PR.

**Tests added (8 new):**
- `TestBraceInStringCatalogParser` (unbalanced + helper unit)
- `TestDepthFlexibleSkillMatch` (depth-5 + depth-4 regression)
- `TestSourcesYamlRootOnly` (non-root rejected, root accepted)
- `TestFileCategoriesDeterminism` (sorted, order-independent)

**54 tests passing (was 46).**

**NITs deferred** (documented for follow-up):
- `dependabot.yml` / `CODEOWNERS` / `LICENSE` land as `unknown` — could add a `config` contribution type if signal becomes noisy
- CLI lacks file-size guard on `--stdin` / `--file` — CI input is `git diff` output, low risk
- Snapshot fixture name "regression for #823" is somewhat misleading — the real PR #823 included a SKILL.md change, so the fixture correctly reflects what the classifier sees but isn't the doc-only regression case. The synthetic `test_pr_823_doc_only_change_does_not_drag_pack_skills` test IS that case.

Jeremy Longshore
intentsolutions.io

PR #856, opened 2026-06-11
feat(ccp): frontmatter-scope kernel-shadow deviation + re-pin @intentsolutions/core 0.4.1

jeremylongshore:
## Scope finding

The kernel-shadow (`scripts/kernel-shadow-validation.mjs`, merged in #855) compared the kernel's authoring/v1 `skill-frontmatter` machine-spec against the existing prose-spec validator (`scripts/validate-skills-schema.py --marketplace --json`). The **raw** per-file PASS/FAIL deviation **conflated two different scopes**:

- The **kernel** `skill-frontmatter` schema is a **FRONTMATTER contract** — it grades the YAML frontmatter block and nothing else.
- The **prose-spec validator** lints **frontmatter AND the markdown body** (required `## Overview` / `## Instructions` / `## Output` / … sections at marketplace tier; body errors are prefixed `[body]`).

So a file with valid frontmatter but missing body sections is **existing-FAIL** (the prose-spec fails it on `[body]` errors) yet **kernel-PASS** (the frontmatter is genuinely valid). That is **not** a kernel gap — it's a **scope difference**. Counting it as a deviation inflated the raw number.

## What this PR does

Scopes the existing-validator verdict to **frontmatter** (the kernel's actual scope) and emits **both** signals in the console summary, `report.json`, the GitHub Step Summary, and the file header docstring:

- **`frontmatterDeviationPct`** — the **cutover-relevant** signal. Every `[body]`-prefixed error is filtered out before counting; a file is a frontmatter deviation only when the kernel and the frontmatter-scoped prose-spec verdict disagree.
- **`bodyScopeOnlyCount`** — **informational**. Files the kernel passes that the prose-spec fails on **body sections only** — a scope difference, NOT a kernel gap.

Mechanism: the batch `--json` run emits only per-file error **counts**; the **single-file** `--json` invocation emits the error **string** list. The shadow runs the batch once to enumerate the corpus + raw verdict, then re-fetches strings per-file **only for the (small) raw-disagreement set** to scope it — never for the whole corpus.

## Re-pin

`@intentsolutions/core` `0.4.0` → exact **`0.4.1`** across `package.json` + `pnpm-lock.yaml` (`pnpm install`) + the script's `KERNEL_PIN` constant + the header docstring. The exact-pin drift-guard still fires on any mismatch.

## Measured frontmatter deviation (real corpus, 3625 files)

| signal | value |
|---|---|
| **frontmatter agree** | **3583** |
| **frontmatter disagree** | **42** |
| **frontmatter deviation rate** | **1.16%** |
| body-scope-only (informational) | 1897 |
| raw disagree (scope-conflated) | 1939 |
| raw deviation rate | 53.49% |
| existing-PASS / kernel-FAIL | 0 |
| existing-FAIL / kernel-PASS | 1939 |

**The frontmatter deviation is 1.16%, NOT ~0%** (the "40/40 `[body]`-only" finding in the task brief was a sample that missed these). Scoping the deviation surfaced **42 genuine frontmatter disagreements** — real prose-spec frontmatter checks the kernel's `skill-frontmatter` schema does **not yet encode**. These are kernel gaps to surface, broken down as:

- **33** — `[frontmatter]` unscoped `Bash` in `allowed-tools` (prose-spec requires scoped `Bash(git:*)` etc; the kernel accepts plain `Bash`). 1× notion-pack, 1× supabase-pack, 1× framecraft, 30× openrouter-pack.
- **7** — `[security]` `description` contains shell-substitution (`$(...)` / backticks / `${VAR}`): 5× langchain-py-pack, 1× podium-pack, 1× penetration-tester.
- **1** — `[frontmatter]` `description` exceeds 1024 characters: `plugins/community/contributing-clanker/skills/contribute/SKILL.md`.
- **1** — `[frontmatter]` `name` contains reserved word: `plugins/community/claude-reflect/SKILL.md`.

(Full list with per-file errors in the `frontmatter-deviations` listing of `scripts/.kernel-shadow/report.json`. Zero scope-fetch failures.)

## Advisory / untouched

- **Advisory · non-blocking · non-required preserved** — no change to `.github/workflows/kernel-shadow-validation.yml` (`continue-on-error: true`, not in the branch-protection required set).
- **Untouched:** `scripts/validate-skills-schema.py`, `.github/workflows/validate-plugins.yml`, and the working-tree `CLAUDE.md` change (deliberately left unstaged).
- Known cosmetic nit (not fixed here, to honor "do not modify any other workflow"): `kernel-shadow-validation.yml` header comments still say `0.4.0` descriptively; the workflow installs the exact pin from the lockfile (now 0.4.1), so behavior is correct.

**Beads:** bd_000-projects-26ef

### DIFF
--- a/package.json
+++ b/package.json
@@ -47,7 +47,7 @@
     "@commitlint/config-conventional": "^20.5.3",
     "@eslint/js": "^9.39.4",
     "@intentsolutions/audit-harness": "^1.1.5",
-    "@intentsolutions/core": "0.4.0",
+    "@intentsolutions/core": "0.4.1",
     "@types/node": "^20.19.41",
     "ajv": "^8.17.1",
     "ajv-formats": "^3.0.1",
--- a/pnpm-lock.yaml
+++ b/pnpm-lock.yaml
@@ -26,8 +26,8 @@ importers:
         specifier: ^1.1.5
         version: 1.1.5
       '@intentsolutions/core':
-        specifier: 0.4.0
-        version: 0.4.0
+        specifier: 0.4.1
+        version: 0.4.1
       '@types/node':
         specifier: ^20.19.41
         version: 20.19.41
@@ -1754,8 +1754,8 @@ packages:
     engines: {node: '>=18'}
     hasBin: true
 
-  '@intentsolutions/core@0.4.0':
-    resolution: {integrity: sha512-rwguEa/l7CDBvrhUskcodZfAs7k0VvkjgcYDc4lQ/L1KNYXroEvwkhYy8PNajlAg4CN6Cb0Tf9fH6jt07B5+Sw==}
+  '@intentsolutions/core@0.4.1':
+    resolution: {integrity: sha512-wfZuLlWmzJ1kuOb7MdxAcmILqzBE/6nz0qJLlCsHjD27Gdv2B+GwPloCXG7He4YKmJlmgsk2Dhfo4iXQcVo/hg==}
     engines: {node: '>=20.0.0', pnpm: '>=9.0.0'}
 
   '@isaacs/cliui@8.0.2':
@@ -6478,7 +6478,7 @@ snapshots:
 
   '@intentsolutions/audit-harness@1.1.5': {}
 
-  '@intentsolutions/core@0.4.0':
+  '@intentsolutions/core@0.4.1':
     dependencies:
       zod: 4.4.3
 
--- a/scripts/kernel-shadow-validation.mjs
+++ b/scripts/kernel-shadow-validation.mjs
@@ -7,7 +7,7 @@
  * WHAT THIS IS
  * ------------
  * Runs the kernel's published machine-spec — the authoring/v1 `skill-frontmatter`
- * JSON Schema from @intentsolutions/core@0.4.0 — over the SAME SKILL.md corpus the
+ * JSON Schema from @intentsolutions/core@0.4.1 — over the SAME SKILL.md corpus the
  * existing prose-spec validator (scripts/validate-skills-schema.py) already grades,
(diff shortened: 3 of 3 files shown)


jeremylongshore:
### CI status: 20 pass · 2 pre-existing reds (not from this PR)

The two failing checks are **pre-existing on `main`** and unrelated to this diff (which changes only `package.json`, `pnpm-lock.yaml`, and `scripts/kernel-shadow-validation.mjs` — zero Python, zero markdown):

- **`markdownlint`** — 3× `MD012/no-multiple-blanks` in `marketplace/src/content/blog-posts/{hitl-delivery-is-a-fail-closed-exactly-once-problem,honor-the-gate-when-the-verdict-is-inconvenient,stop-crying-wolf-3-strike-uptime-monitor-gate}.md`. These blog files exist unchanged on `main` (authored in a prior blog-mirror commit); not touched here.
- **`test (python-tests)`** — 6 failures all in `scripts/pr-prescreen/test_summarize.py` (`module 'summarize' has no attribute 'GROQ_URL'`, `'skipped: no api key' != 'ok'`, `KeyError: 'body'`, …) — environment/mock-dependent pr-prescreen summarizer tests. `validate-skills-schema.py` is untouched.

Both reds reproduce on the base commit (`22f99a9`) and the latest `main` runs are already `failure`, so they are out of scope for this kernel-shadow change. Fixing them here would be unrelated scope creep.

The check this PR is actually about — **`kernel-shadow (advisory, non-blocking)`** — **passes** (3m48s), running the refined frontmatter-scoped shadow against the re-pinned `@intentsolutions/core@0.4.1`.

- Jeremy Longshore
intentsolutions.io

PR #888, opened 2026-06-20
test(kernel-vendor-hash): formalize V<=C<=K ordering invariant test corpus

jeremylongshore:
## What

Adds an executable fixture corpus that formalizes the version-coupling ordering invariant **V ≤ C ≤ K** + bounded-staleness model (plan 033 § 14.15.1) against the existing `kernel-vendor-hash` comparator. Closes F-MK-003.

The three version identities:
- **V** vendored snapshot · **C** CCPI-declared kernel pin · **K** kernel latest published.

### Changes

- **Surgical refactor of `scripts/kernel-vendor-hash.mjs`** — extract the pure V≤C≤K + staleness evaluation into an exported `evaluateCoupling()` that `main()` now delegates to (no behavior change; CLI output verified identical before/after), export `semverCompare`, and guard the `main()` call so the module is import-safe (runs only when invoked as a CLI, never on import). This keeps the CLI and the tests on a single code path — no logic fork.
- **New `scripts/kernel-vendor-hash.test.mjs`** (node's built-in `node --test` runner, **zero new dependencies**), 20 tests:
  - **Positive fixtures** (the bead's a/b/c): (a) vendored=CCP=kernel; (b) vendored<CCP=kernel within the 7-day budget; (c) vendored=CCP<kernel (kernel ahead).
  - **Negative fixtures**: vendored>CCP, CCP>kernel, vendored>kernel (both ordering legs), and a >7-day staleness breach — each must report a violation/stale.
  - **Edge cases**: staleness boundary (7d in-budget vs 8d over), unknown kernel age, existence guards (absent V/C/K warn + skip, never crash/violate), ranged-pin warning.
  - **CLI exit-code contract** (end-to-end subprocess): advisory exits 0 even on a violation; `--strict` exits 1 on a real ordering violation; clean ordering exits 0.
- **Wire a BLOCKING test step into `.github/workflows/kernel-vendor-hash.yml`** + add the test file to the path filter.

## Why

The version-coupling check shipped advisory-only with no test corpus. Comparator **correctness** is orthogonal to the DR-049 advisory soak (which governs only pin-bump pressure / validator authority) — a broken comparator is a real bug. This pins the ordering invariant's behavior so a regression turns CI red, while the check itself stays advisory on pin-bump pressure during the soak.

## Verification

All green:
- `node --test scripts/kernel-vendor-hash.test.mjs` → **20 pass / 0 fail**
- `node scripts/kernel-vendor-hash.mjs` (advisory smoke) → exit 0
- `eslint` + `prettier` (`scripts/**`) clean
- `audit-harness verify` → `harness-hash: OK`
- Full pre-commit gate (lint-staged + harness) passed on commit

- Jeremy Longshore
intentsolutions.io

## Summary by CodeRabbit

* **Tests**
  * Added a full Node test suite covering semver ordering, coupling validation, and staleness boundary behavior (including disabled/stale/warning scenarios).
  * Added end-to-end CLI checks for both advisory and strict exit-code contracts.
* **New Features**
  * Added blocking CI validation that runs the coupling test corpus before the advisory reporting step.
* **Refactor**
  * Refactored the kernel coupling evaluation to be reusable and parameterized, while keeping the existing report output behavior.
* **Chores**
  * Updated the CI workflow triggers so changes to the test corpus re-run the validation workflow.

### DIFF
--- a/.github/workflows/kernel-vendor-hash.yml
+++ b/.github/workflows/kernel-vendor-hash.yml
@@ -33,6 +33,7 @@ on:
   pull_request:
     paths:
       - 'scripts/kernel-vendor-hash.mjs'
+      - 'scripts/kernel-vendor-hash.test.mjs'
       - 'scripts/validate-skills-schema.py'
       - 'package.json'
       - 'pnpm-lock.yaml'
@@ -71,6 +72,16 @@ jobs:
       # V from node_modules in the absence of a .kernel-vendor/ snapshot.
       - name: Install root dev dependencies
         run: pnpm install --frozen-lockfile --filter "claude-code-plugins-monorepo"
+
+      # BLOCKING — the version-ordering invariant test corpus (F-MK-003). This is
+      # NOT advisory: it exercises the V <= C <= K comparator (`evaluateCoupling` +
+      # `semverCompare`) with synthetic positive/negative/staleness fixtures and the
+      # CLI advisory-vs-strict exit-code contract. Correctness of the comparator is
+      # ORTHOGONAL to the DR-049 soak (which only governs pin-bump pressure / validator
+      # authority) — a broken comparator is a real bug, so a failing corpus turns this
+      # check red. Uses node's built-in runner (zero new deps).
+      - name: Version-ordering invariant test corpus (blocking)
+        run: node --test scripts/kernel-vendor-hash.test.mjs
 
       # K (kernel latest published) is resolved from npm in CI — the read-only
       # sibling repo is not checked out here. Pass it explicitly so C <= K and the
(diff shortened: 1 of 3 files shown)


jeremylongshore:
Addressed in 94ca3e205. The CLI smokes now derive the declared pin (C) from a baseline advisory `--json` run via `resolveDeclaredC()` and build the violation regex from it, and force the C>K case with a guaranteed-low K (`0.0.0`) instead of a literal that depends on the current pin. A routine `package.json` pin bump no longer breaks the blocking suite. 20/20 tests pass; eslint + prettier clean.

- Jeremy Longshore
intentsolutions.io

PR #890, opened 2026-06-22
Add mytradeledger-skills: MyTradeLedger REST API skill

trpsbill:
Adds `mytradeledger-skills` to the community marketplace via **Path B (external sync)** — a single `sources.yaml` entry pointing at the source repo, which stays the source of truth. No vendored content.

### What changed (revised per review)
This PR was reworked from a hand-vendored plugin shell to a pure sync source. The diff is now **one file, `sources.yaml`** (+28 lines):

- New `mytradeledger-skills` source → `repo: trpsbill/skills`, `target_path: plugins/community/mytradeledger-skills`, `category: community`
- Syncs `skills/**` (the real `SKILL.md` + `scripts/mtl.py`), `.claude-plugin/plugin.json`, and `LICENSE`
- The weekly external-sync workflow generates the plugin dir, `package.json`, README, and the `marketplace.extended.json` / `marketplace.json` catalog entries — so none of those are hand-edited here

Reverted from the previous revision: the empty `plugins/community/mytradeledger-skills/` shell (`plugin.json`/`README.md`/`package.json`) and the manual `marketplace.*` + root `README.md` edits.

### Source of truth
`mytradeledger` wraps the MyTradeLedger REST API: ledger entries (CRUD + bulk-import + P&L recompute), account balances and realized P&L, CSV export, asset registry, entry metadata, and PAT management. Stdlib-only Python helper, auth via a local `.env`. Lives in https://github.com/trpsbill/skills and stays maintained there.


jeremylongshore:
Thanks for submitting mytradeledger-skills. I did a first-review pass, and there's a structural blocker: **this PR adds an empty plugin shell.** `plugins/community/mytradeledger-skills/` contains only `plugin.json` + `README.md` + `package.json` — **no `skills/`, `commands/`, or `agents/`**, so there's nothing to install. The plugin ships zero capability, and the README actually points users to your *own* marketplace (`/plugin install mytradeledger-skills@trpsbill-skills`).

The MyTradeLedger concept is fine — we just need the real content (or the wiring to fetch it). Two supported paths:

**Path B — sync from your repo (recommended here).** Since you maintain the real plugin in [`trpsbill/skills`](https://github.com/trpsbill/skills), add a single entry to [`sources.yaml`](https://github.com/jeremylongshore/claude-code-plugins-plus-skills/blob/main/sources.yaml) pointing at it. The weekly sync then pulls your actual `SKILL.md` content into `plugins/community/mytradeledger-skills/` and your repo stays the source of truth. This is the right path for an externally-maintained plugin.

**Path A — vendor the real content.** Include the actual `skills/<name>/SKILL.md` file(s) with the 8-field marketplace frontmatter in this PR — not just the shell.

Minor (flagged by the bot): the `package.json` install handle doesn't match the README; worth aligning once you pick a path.

Happy to merge once there's real skill content (or a `sources.yaml` entry that fetches it). Let me know which path you want.

- Jeremy Longshore
intentsolutions.io

trpsbill:
Thanks for the review — went with **Path B** as recommended.

I dropped the empty plugin shell and the manual `marketplace.*` / README edits entirely. The PR is now a single `sources.yaml` entry that points the weekly sync at `trpsbill/skills` (syncing `skills/**` → the real `SKILL.md` + `scripts/mtl.py`, plus `.claude-plugin/plugin.json` and `LICENSE`). The sync then generates `plugins/community/mytradeledger-skills/`, the `package.json`, the README, and the catalog entries — so my repo stays the source of truth and there's nothing hand-vendored to drift.

On the bot's minor note (package.json install handle vs README): that mismatch is gone since the vendored `package.json` was removed — the install handle is now whatever the sync generates from `name: mytradeledger-skills`.

Diff is down to one file (+28 lines). Let me know if you'd like the source synced as a flattened single-skill (`source_path: skills/mytradeledger`) instead of the whole-plugin layout, or if `plugins/crypto/` would be a better home than `plugins/community/`.

PR #911, opened 2026-06-28
fix(validator): accept current GA plugin.json manifest fields (schema 3.12.0)

jeremylongshore:
## Problem

The `plugin.json` allowlist (`PLUGIN_JSON_FIELDS` in `scripts/validate-skills-schema.py`) carried only the **pre-GA 15** fields. `validate_plugin_json()` then appends any other key as an **ERROR** (`Unknown field: '<x>' — not in Anthropic spec`). The Anthropic plugin manifest has since gained several **GA** fields — so the validator now **hard-rejects valid, current Anthropic plugins**. This is the biggest gap in our plugin-manifest validation.

Verified against the official spec: [code.claude.com/docs/en/plugins-reference](https://code.claude.com/docs/en/plugins-reference) § "Plugin manifest schema".

## Fix (additive, spec-compliance)

Add to `PLUGIN_JSON_FIELDS` (all optional; `name` stays the only required field):

| Field | Type | Status |
|---|---|---|
| `displayName` | string | GA (Claude Code v2.1.143+) |
| `defaultEnabled` | boolean | GA (v2.1.154+) |
| `dependencies` | array | GA (plugin deps w/ semver) |
| `userConfig` | object | GA |
| `channels` | array | GA |
| `$schema` | string | GA (editor metadata, ignored at load) |
| `experimental` | object | experimental (`experimental.{themes,monitors}`) |

- `TYPE_MAP` gains `boolean` (for `defaultEnabled`).
- `SCHEMA_VERSION` 3.11.0 → **3.12.0**; `SCHEMA_CHANGELOG.md` entry added.
- `/validate-plugin` reference snapshot (`plugin-schema.md`) synced — `displayName` removed from the "invalid fields" list, field reference now lists all 22.
- New test `tests/test_validate_plugin_json_ga_fields.py` pins GA-field acceptance, boolean typing, name-still-required, and that genuinely-unknown fields are still flagged.

**Purely additive** — accepts more, rejects nothing previously accepted, so **no plugin can regress**. Existing 26 frontmatter tests + 6 new tests pass.

Per `SCHEMA_CHANGELOG.md` **NON-NEGOTIABLE #6** ("adding missing documented fields" — autonomous-OK).

## ⚠️ Decision needed — NON-NEGOTIABLE #7 (NOT done here)

The official `claude plugin validate` treats **unrecognized** fields as **warnings** (errors only under `--strict`). Our validator still **errors** on genuinely-unknown fields. Aligning to Anthropic's warn-not-error policy is an **error-vs-warning-semantics** change — **NON-NEGOTIABLE #7, needs your explicit sign-off**, so it is **deferred**.

Adding the GA fields above already removes the false rejection of valid current plugins (those fields are no longer "unknown"). The remaining decision only affects typos / other-ecosystem metadata. Two options if/when you want it:

1. Keep erroring (status quo) — strictest; catches typos but diverges from Anthropic.
2. Warn-not-error to match Anthropic, with an opt-in `--strict` for CI.

Say the word and I'll do (2) in a follow-up.

- Jeremy Longshore
intentsolutions.io

### DIFF
--- a/000-docs/SCHEMA_CHANGELOG.md
+++ b/000-docs/SCHEMA_CHANGELOG.md
@@ -146,6 +146,45 @@ compliance are welcome; structural changes to the IS rubric are not.
 
 ---
 
+## [3.12.0] — 2026-06-28 — plugin.json manifest brought up to current Anthropic GA spec (additive)
+
+**Additive MINOR — no change to any required-field set, tier model, or
+error-vs-warning semantics.** `PLUGIN_JSON_FIELDS` (the `.claude-plugin/plugin.json`
+manifest allowlist) had only the pre-GA 15 fields. The Anthropic plugin manifest
+has since gained several **GA** fields; the validator was therefore hard-rejecting
+valid, current Anthropic plugins with `Unknown field: '<x>' — not in Anthropic
+spec` (an ERROR), which is no longer true for these fields.
+
+Added to `PLUGIN_JSON_FIELDS` (all optional; `name` remains the only required
+field, unchanged):
+
+| Field | Type | Status | Note |
+|---|---|---|---|
+| `displayName` | string | GA (Claude Code v2.1.143+) | human-readable name in the `/plugin` picker |
+| `defaultEnabled` | boolean | GA (v2.1.154+) | enabled by default absent a user preference |
+| `dependencies` | array | GA | required plugins w/ optional semver constraints |
+| `userConfig` | object | GA | user-configurable values prompted at enable time |
+| `channels` | array | GA | message-channel declarations bound to MCP servers |
+| `$schema` | string | GA (metadata only) | JSON Schema URL; ignored at load time |
+| `experimental` | object | experimental | holds `experimental.themes` / `experimental.monitors` |
+
+`TYPE_MAP` in `validate_plugin_json()` gained `boolean` (for `defaultEnabled`).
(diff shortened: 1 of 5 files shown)


jeremylongshore:
**NON-NEGOTIABLE #7 sign-off (audit trail).** Per `000-docs/SCHEMA_CHANGELOG.md` NON-NEGOTIABLE #7, the error-vs-warning change requires explicit approval recorded in the PR thread before it lands. Jeremy Longshore approved the flip on 2026-06-28.

Implemented in `3914b1b` (schema **3.13.0**), folded into this PR on top of the GA-fields add:

- Unrecognized plugin.json field → **WARNING** by default (was ERROR), matching Anthropic's own `claude plugin validate`.
- New **`--strict`** flag promotes them back to errors (CI opt-in); verified end-to-end (default → WARN, `--strict` → ERROR).
- Wrong-**type** field + missing required **`name`** stay errors regardless of `--strict`.
- `difflib` "did you mean …" near-miss hint.
- No change to any required-fields set or tier (NON-NEGOTIABLES #1/#2 intact).

This also makes plugin.json consistent with the SKILL.md + agent validators, which already warn (never error) on unknown fields. 9 plugin + 26 frontmatter tests green.

- Jeremy Longshore
intentsolutions.io

PR #914, opened 2026-06-29
feat(validator): L2 consumer-cutover — drift-gate PLUGIN_JSON_FIELDS against the kernel (schema 3.14.0)

jeremylongshore:
**Closes the second half of the auto-currency loop** (692-AA-AACR § 0.5). L1 made the kernel SSoT durably current; L2 makes CCPI's plugin validator *consume* it instead of being hand-maintained.

`PLUGIN_JSON_FIELDS` had **no** comparison to the kernel, so an upstream GA-field change silently went stale and required a hand-patch (the 3.12.0 event). Now:

- **`load_kernel_plugin_fields()`** reads the kernel's **current** plugin-manifest surface from `@intentsolutions/core` `authoring/`**`v2`** (`upstream-base ∪ is-overlay`). v2 not v1 — v1 is an early ~10-field capture; v2 carries the full GA surface. Needs core `>=0.9.0` (also reconciled the stale `0.4.1` node_modules install via `pnpm install`).
- **`kernel_shadow_report()`** gains a `plugin_manifest` block; **`print_kernel_shadow()`** WARNs when `PLUGIN_JSON_FIELDS` is stale. The existing **`kernel-shadow-validation.yml`** CI lane now surfaces plugin-manifest drift automatically.
- The shadow **immediately caught a drift**: the kernel sanctions a `metadata` plugin.json field (v2 is-overlay) CCPI lacked → adopted into `PLUGIN_JSON_FIELDS` (now `fields_match: true`, 23 fields).

**Advisory** — `PLUGIN_JSON_FIELDS` stays authoritative until the DR-049 flip; same posture as the skill-frontmatter + agent shadows. No NON-NEGOTIABLE touched. 2 new tests; suite green (11 plugin + 26 frontmatter).

- Jeremy Longshore
intentsolutions.io

### DIFF
--- a/000-docs/692-AA-AACR-spec-currency-audit-2026-06-28.md
+++ b/000-docs/692-AA-AACR-spec-currency-audit-2026-06-28.md
@@ -1,0 +1,230 @@
+# Spec-Currency Audit — Claude Code authoring artifacts vs IS validators (2026-06-28)
+
+**Code:** 692-AA-AACR · **Author:** Jeremy Longshore · **Date:** 2026-06-28
+**Scope:** Are the Intent Solutions validators (`scripts/validate-skills-schema.py`,
+the `/validate-*` skills, the catalog scripts, the vendored kernel schemas) current
+against the live Anthropic Claude Code authoring specs as of mid-2026?
+**Method:** Six authoring surfaces fetched + cited from `code.claude.com/docs`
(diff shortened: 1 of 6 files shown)


jeremylongshore:
Superseded by a clean single-commit L2 PR (local main had diverged from origin's squash, pulling in extra commits).

PR #916, opened 2026-06-29
fix(bash): safe empty-array expansion for bash 3.2 (macOS)

olispeedy:
## Problem

On macOS, the default shell is **bash 3.2**. When `set -u` is active, expanding an empty array raises an *unbound variable* error:

```bash
# fails on bash 3.2 when EXCLUDE_PATHS is empty
ALL_EXCLUDES=("${DEFAULT_EXCLUDES[@]}" "${EXCLUDE_PATHS[@]}")
# bash: EXCLUDE_PATHS[@]: unbound variable
```

This causes the hook to crash on a stock macOS system whenever no custom `excludePaths` are configured.

## Fix

Use the conditional expansion idiom `${var[@]+"${var[@]}"}`, which expands to nothing when the array is unset or empty and is safe across all bash versions:

```bash
ALL_EXCLUDES=("${DEFAULT_EXCLUDES[@]}" ${EXCLUDE_PATHS[@]+"${EXCLUDE_PATHS[@]}"})
```

No behaviour change when `EXCLUDE_PATHS` has elements.

## Tested on

- macOS bash 3.2.57 (stock macOS shell)
- bash 5.x (Homebrew)

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/plugins/productivity/prettier-markdown-hook/scripts/format-markdown.sh
+++ b/plugins/productivity/prettier-markdown-hook/scripts/format-markdown.sh
@@ -142,7 +142,9 @@ load_config() {
     fi
 
     # Combine default + custom exclusions (additive pattern)
-    ALL_EXCLUDES=("${DEFAULT_EXCLUDES[@]}" "${EXCLUDE_PATHS[@]}")
+    # Use conditional expansion to avoid unbound variable error on bash 3.2 (macOS default)
+    # when EXCLUDE_PATHS is empty and set -u is active.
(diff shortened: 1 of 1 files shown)


olispeedy:
Good catch by @greptile-apps — the original fix left the outer conditional unquoted, which would word-split paths containing spaces when appended to the array.

Updated to wrap the entire conditional expansion in double quotes:

\`\`\`bash
"${EXCLUDE_PATHS[@]+"${EXCLUDE_PATHS[@]}"}"
\`\`\`

Tested locally on bash 3.2.57 (stock macOS):
- Empty \`EXCLUDE_PATHS\` — no crash ✓
- Paths with spaces (e.g. \`"my project"\`) — preserved as single elements ✓

Issue #927, opened 2026-07-01
Bump jrig-cli pin to 0.1.2 and harden validate-plugins CI (deadline gate, sync-lint drift, npm-audit policy, forge_proofs)


jeremylongshore:
Cross-repo remediation cluster — the cc-plugins-anchored half. Executes source plan `intent-eval-lab/000-docs/106-PP-PLAN-cicd-jrig-cross-repo-remediation-2026-07-01.md` (full send A+B+C+D). Sibling cluster (the j-rig release + eval instrumentation) is tracked on `jeremylongshore/j-rig-skill-binary-eval`.

### Scope (one issue per logical cluster, not per task)

- **A2 — bump the jrig-cli pin `0.1.1 → 0.1.2`** (`package.json`) + refresh the lockfile. Gated on the `@intentsolutions/jrig-cli@0.1.2` publish propagating to npm. Restores the #173 + #175 provider/judge fixes to skill evals (fixes the Xquik false NO-SHIP, `693-AA-AACR`).
- **A3 — populate the first `forge_proofs` tier3-jrig row + make `enrich-jrig-data.mjs` fail loudly on a missing DB.** Today `marketplace/src/data/jrig-data.json` is literally `{}`, so every JRig-Verified badge is "pending" and a missing DB degrades silently.
- **B1 — register a dated deadline gate for the frontmatter `--strict` campaign** and correct the live error count (live `validate --strict` = **317** errors, not the stale 171/182). NOT the `|| exit 1` flip — that's a follow-on once the campaign clears. `scripts/check-ci-deadlines.py` already runs in CI, so the `# REPORT-ONLY-UNTIL:` marker is a real forcing function.
- **B2 — add the `sync-lint-ignores` generator + `--check` drift gate** and close the 2 sync-external blockers from `000-docs/691-AT-AUDT` (markdownlint/ruff synced-dir exclusion drift; `synced-gitignored-files-dropped`). Remaining 4 medium/low sync bugs → opportunistic follow-ups.
- **B3 — decide + implement the npm-audit CI policy** (`npm audit --production || true` at `validate-plugins.yml:284`): blocking on high/critical, or tracked report-only with a `check-ci-deadlines.py` deadline.

### Guardrails

No rule-loosening, no threshold-lowering, no coverage exclusions. Fixes land on the artifact, or a gate is honestly reclassified with a documented reason. B1 is explicitly **not** the strict-flip this pass.

[...]

PR #931, opened 2026-07-01
ci: derive synced-plugin lint exclusions with a generator (self-registering + drift-gated)

jeremylongshore:
The synced-plugin lint-exclusion lists in `.markdownlint-cli2.jsonc` (`ignores`) and `ruff.toml` (`extend-exclude`) were **hand-maintained**, and the CI gate (`check-synced-lint-exclusions.mjs`) could only **report** drift — a human still had to add the lines by hand. A newly-synced source therefore red-failed the very next CI run on mirrored markdown/python it doesn't own (the bug that blocked the `beads-dolt` sync, per `000-docs/691-AT-AUDT`).

Replaces the check-only script with a **generator**, `scripts/sync-lint-ignores.mjs`:

- **WRITE mode** derives both lists from the `.source.json`-marked dirs (identical discovery to the old check) and canonicalizes a **sentinel-delimited managed block** in each config. Everything outside the `BEGIN`/`END` sentinels (manual entries, comments, array close) is preserved — hand-maintained entries can never be clobbered (the old script's safety property is kept).
- **`--check` mode** is the CI drift gate: exits 1 naming the missing/stale dirs + the fix command. `validate-plugins.yml` repointed at it.
- Wired **WRITE mode into `sync-external.yml`** right after the sync writes files and before change-detection/commit, so a newly-synced source **self-registers** its exclusion in the same auto-PR (guarded on non-dry-run + zero sync errors, matching downstream commit gating). This unblocks re-managing `beads-dolt` as a normal external sync.

`markdownlint-cli2` has no real `--ignore` flag, so the config arrays are the only authoritative exclusion surface; the generator keeps them derived. Deletes the superseded `scripts/check-synced-lint-exclusions.mjs` and updates `ruff.toml`'s comment reference.

### Verified
- `--check` exits 0 (51 synced dirs in sync); WRITE mode idempotent.
- `ruff.toml` valid TOML (56 excludes) + `.markdownlint-cli2.jsonc` valid JSONC (62 ignores); both workflows valid YAML.
- The only reorder is ruff's out-of-order `publishing-skills` sorting into canonical place.

Part of #927. (One of the 2 sync-external blockers from `691-AT-AUDT`; the `synced-gitignored-files-dropped` blocker is tracked separately.)

**Beads:** `bd_000-projects-ueeqe` — under epic `bd_000-projects-4yxwc`.

- Jeremy Longshore
intentsolutions.io


jeremylongshore:
Good catch — fixed in `087de818a`. The generator now detects the file's existing EOL (`\r\n` vs `\n`) and splits/joins with it, so a CRLF checkout (Windows / `core.autocrlf`) no longer strands a trailing `\r` that breaks the sentinel lookup. `--check` still green (51 synced dirs in sync).

- Jeremy Longshore
intentsolutions.io

Issue #966, opened 2026-07-07
Harden the external plugin sync against supply-chain injection (lockfile pinning + security scan + surface tiers)


jeremylongshore:
## The gap

The weekly external sync (`sync-external.yml` → `scripts/sync-external.mjs`) mirrors whatever each upstream's default branch says at cron time:

- `sources.yaml` pins **nothing** — no commit, no content hash. `verified: true` vets a source once, at listing; every subsequent upstream change rides in automatically.
- The mirrored content is **not just markdown**: `plugins/community/` currently holds 67 shell scripts, 11 Python files, 4 JS files, and 52 JSONs (hook/MCP surfaces execute automatically in users' agents).
- No deterministic security scan runs on sync PRs. The gate today is a human checklist + AI reviewers over a multi-plugin weekly diff.

**The attack:** a listed author's account is compromised (or goes rogue) → upstream gains an exfil instruction or a poisoned script → Monday's sync PR mirrors it inside a routine diff → merged → every marketplace user's agent executes it with the user's credentials and tools.

## The fix (three layers)

1. **Lockfile pinning** — `sources.lock.json`: per-file sha256 recorded at vet time. Sync compares before mirroring; unchanged content flows into the routine PR; **any drift quarantines that source into its own review PR** with a per-source diff. Lock updates are the approval record. Bootstrap: current in-tree content (already human-merged) seeds the baseline.
2. **Deterministic security scan** — REFUSE/CHALLENGE/FLAG pattern gate (pipe-to-shell, encoded exec, env/secret exfiltration, network beacons in scripts, hook/MCP additions, `allowed-tools` expansion, obfuscation smells) as a hard CI check on any PR touching `plugins/community/**` or `sources.yaml`, and as a step inside the sync itself.
3. **Surface tiers by blast radius** — `markdown-only` / `scripted` / `hooks-mcp` per source; the heavy tiers require curated/frozen handling where every upstream change needs explicit human re-approval.

**Beads:** claude-2xxr (epic) + children .1–.3

- Jeremy Longshore
intentsolutions.io

jeremylongshore:
Linked bead: `claude-2xxr` (source of truth). Use `bd-sync` to mirror updates.

jeremylongshore:
**bd-sync from `claude-2xxr`:** Child .3 (threat model + vetting playbook) landed: PR #967 merged. 1/3 epic children complete.

jeremylongshore:
[...]

PR #967, opened 2026-07-07
docs(security): external-sync threat model + contributor/maintainer vetting playbook

jeremylongshore:
## What (Track C of #966 — docs only)

The weekly external sync mirrors 54 unpinned upstream repos into `plugins/` — including 67 shell scripts, 11 Python files, and hook/MCP configs that execute automatically in users' agents. This PR files the two documents that anchor the hardening story:

| Doc | What it covers |
|---|---|
| `000-docs/698-TQ-SECU-external-sync-threat-model.md` | Assets (users' agents + credentials, marketplace trust), the trust boundary (upstreams we do not control), six attack vectors (unpinned rug-pull, compromised author account, poisoned script/hook/MCP, prompt injection in markdown, typosquat/dependency confusion, sync-machinery abuse), the mitigation layers stacked per vector, and an honest residual-risk section (content-not-intent, pattern-scanner limits, review fatigue, trusted-by-baseline bootstrap, no revocation push). |
| `000-docs/699-DR-GUID-external-source-vetting-playbook.md` | The human half: what `verified:` and `curated:` actually gate, the 9-step listing checklist before a source enters `sources.yaml`, the lockfile drift-quarantine review procedure, CHALLENGE sign-off via scoped scan-allowlist entries, the `markdown-only` / `scripted` / `hooks-mcp` surface-tier policy, source suspension/removal, and anti-fatigue review discipline. |

`SECURITY.md` gains a new "External Plugin Sync — Supply-Chain Security" section linking both docs — **additive only**, no existing content changed.

## Placement

Filed in `000-docs/` per the repo's `NNN-CC-ABCD` doc-filing convention (next sequence numbers 698/699; `TQ-SECU` and `DR-GUID` codes already in use here). Force-added past the root `000-docs/*` ignore, following the precedent of 695/696/697 (filed doc classes outside the re-include allowlist) — these two must be tracked because SECURITY.md links them.

## One story with the sibling PRs

- `feat/sync-lockfile-pinning` — Track A: per-file sha256 pinning in `sources.lock.json` + drift quarantine PRs (lock update = approval record)
- `feat/sync-security-scan` — Track B: deterministic REFUSE/CHALLENGE/FLAG scan as a hard gate on `plugins/community/**` + `sources.yaml`
- this PR — Track C: the threat model both operate inside, and the human procedures that make them stick

Where this doc set and the merged implementations differ on mechanics, the code is authoritative — both docs say so explicitly.

## Scope extension (documented)

[...]

PR #968, opened 2026-07-07
feat(security): pin external sync sources with a content lockfile + drift quarantine

jeremylongshore:
## Track A — lockfile pinning + drift quarantine

Closes the unpinned-upstream gap in the external plugin sync. Refs #966.

### Threat

`sources.yaml` pins nothing. The weekly cron (`sync-external.yml`, Mondays 06:00 UTC) mirrors each upstream's **live default branch** — including 67 shell scripts, 11 python files, and JSON hook/MCP configs — into `plugins/` via an auto-generated PR. A source vetted once at listing silently absorbs **every later upstream push**: a compromised or rogue upstream author's next commit rides into the marketplace and executes in users' agents.

### Mitigation

**`sources.lock.json`** (repo root) pins, per source: the resolved upstream commit sha (`resolved_ref`), a `sha256:` digest of every mirrored file, and `locked_at`. It is deterministic (sorted source names, sorted file paths, fixed entry key order) — **its git diff is the security review surface.**

On every sync run, the engine diffs the freshly-cloned upstream bytes against the lock **before writing anything**:

| Classification | Behavior |
|---|---|
| `new-source` (no lock entry) | Mirror as today + record the baseline. Vetting-at-listing stays the human gate. |
| `unchanged` | Mirror as today (no-op diffs). Lock untouched, except a one-time backfill of a `null` `resolved_ref` left by the bootstrap. |
| `drifted` | **QUARANTINED** — nothing of that source is mirrored this run. The run goes red exactly like the existing partial-sync path (commit/PR steps are additionally gated on `quarantined == '0'`), and the failed sources are named with per-file added/removed/changed counts. |

A drifted lock only advances via an explicit **`--relock=<source>` / `--relock-all`** (CLI) or the new **`relock` workflow_dispatch input** (comma-separated names or `all`) — a deliberate human approval after reviewing the upstream diff. Notably:

- `--force` does **NOT** bypass the gate (the weekly workflow always passes `--force`; forcing file writes must never double as sight-unseen content approval).
- A `repository_dispatch` payload **cannot** relock — approval is workflow_dispatch-only (Actions UI) or local CLI.

### Fail-closed details

- A corrupt `sources.lock.json` **aborts the run** rather than degrading to "nothing is pinned" (which would re-baseline drifted content).
- An upstream that suddenly yields an empty file set classifies as drifted (all-removed) — which also guards the orphan-prune path from a mass delete.
- `--strict` exits non-zero on any quarantine (in addition to the existing partial-failure behavior).
- The lock advances only **after** a source's mirror writes succeeded — a mid-mirror throw never records an approved baseline for content that never landed.
- Curated-frozen sources (`tonone`, `hyperflow`) never reach the gate (mirror frozen upstream of it); their bootstrapped entries become active drift baselines if `curated:` is ever dropped — resuming the mirror then requires an explicit relock review.

### Bootstrap (this PR's `sources.lock.json`)

Generating from upstream clones would pin **unreviewed** current upstream state, so the initial lock was generated **from the local mirrored files under each source's `target_path`** — the already-human-reviewed baseline — using `.source.json` owned-file `files[]` manifests where present (3 sources) and the engine's own include/exclude glob semantics otherwise, excluding engine-synthesized `README.md` / `.claude-plugin/plugin.json` / `.source.json`. 53 of 54 sources locked; `mytradeledger-skills` has no local mirror yet and will baseline as `new-source` on its first sync. `resolved_ref` is `null` for bootstrapped entries and backfills on each source's first content-unchanged sync.

**Expected rollout behavior:** the first cron run after merge will quarantine every source whose upstream moved since its last mirror (verified live: `gastown` upstream deleted its README + 7 reference files since the last sync — the gate caught it). That first red run is the designed review moment: inspect each named source's upstream diff, then dispatch with `relock=<approved names>`.

### Verification

- `node --test scripts/sync-lockfile.test.mjs` — **17/17 pass**: digest determinism + known sha256 vector, new-source/unchanged/drifted classification with per-file added/removed/changed attribution, drift on a **single flipped byte of one file**, upstream-emptied → drifted, deterministic serialization (insertion order never leaks), fail-closed `loadLock` on corrupt/wrong-shape files, round-trip + no-op re-save.
- Live end-to-end (real upstream clones): `gastown` drifted → quarantined with correct file attribution; `--relock=gastown` (dry-run) → re-baselined; `ejentum-reasoning` → unchanged. `--strict` + quarantine → exit 1; `GITHUB_OUTPUT` carries `quarantined=1`, `quarantined_sources=gastown`.
- New paths-scoped workflow `sync-lockfile-test.yml` runs the corpus on PRs (blocking) and verifies the committed lock parses + stays in canonical deterministic form (hand-edit/merge-mangle detection).
- `actionlint`, `eslint`, `prettier --check` clean on all touched files; `kernel-vendor-hash` corpus still 20/20.

### Residual risk (why the scanner track exists)

The lock pins **content, not intent**: a malicious file already present at bootstrap (or waved through a careless relock) is trusted-by-baseline. The lock converts *silent* upstream absorption into an *explicit, named, diff-reviewable* approval step — it does not judge what the bytes do. Content-level judgment is the scanner track's job; the two gates compose.

- Jeremy Longshore
intentsolutions.io


jeremylongshore:
## Adversarial verify: strong work, but 2 MAJOR fail-open/gap findings before merge

The verifier byte-checked all 1,305 pinned digests against the in-tree mirrors (0 mismatches), confirmed curated sources are correctly skipped, `repository_dispatch` cannot relock, and quarantine returns before any write/prune/lock-advance. Core design is sound. Two MAJORs hold it:

**MAJOR 1 — fail-open: a malformed lock entry classifies as `new-source` and silently re-baselines.** In `diffSource`, `!entry.files || typeof entry.files !== 'object'` → `new-source`, so an entry with `files: null` or a missing `files` key makes the engine mirror current upstream AND record a fresh baseline sight-unseen — the exact "corrupt lock degrades to unpinned" that `loadLock` fails closed against at the JSON layer. Reachable via a merge-mangled or hand-edited lock on main (the canonical-form guard only runs on PRs touching the lock paths). Fix: `diffSource` throws (or classifies `drifted`) when `entry` exists but `entry.files` is not a plain object; `loadLock` rejects a non-object `sources`/entries instead of normalizing. Being applied.

**MAJOR 2 — the corpus test isn't merge-blocking.** `sync-lockfile-test.yml` is a new paths-scoped job, not in main's required-status set — so a change that breaks classification can merge green. Needs adding to branch protection (UI action) once the workflow name is final.

MINORs (tracked): exec-bit flips bypass the content-only pin; the `--strict` comment is aspirational; `resolved_ref` backfill can't land during a quarantined run; no engine-side enforcement test. Rollout note the builder flagged is correct and important: the **first** cron after merge will quarantine every source whose upstream has moved (gastown already drifts) — that red run is the designed review-and-relock moment.

- Jeremy Longshore
intentsolutions.io

jeremylongshore:
Addressed both Gemini review findings (commit `59cbd243f`) — both are fail-closed hardenings of `scripts/sync-lockfile.mjs`:

1. **`loadLock` (was line 106):** a present-but-non-object `sources` field (array/string/null) now **throws** instead of silently defaulting to `{}`. Silent default would classify every pinned source as `new-source` and re-baseline the whole lock — the exact drift-quarantine bypass this file exists to prevent. An *absent* `sources` field still bootstraps to the empty lock.
2. **`diffSource` (was line 172):** only a genuinely **absent** source is `new-source`. A source that is present in the lock but whose entry is corrupt (missing/invalid `files`) now **throws** rather than being laundered into a silent re-baseline.

Added two regression tests (present non-object `sources` rejected; present-but-corrupt entry throws). Full `sync-lockfile` suite 19/19 green; markdownlint/eslint/format all green after rebasing onto the merged #967 (which carried the blog MD012 fix).

- Jeremy Longshore
intentsolutions.io

PR #969, opened 2026-07-07
feat(security): deterministic supply-chain scanner + CI gate for synced content

jeremylongshore:
## What & why

The weekly `sync-external` cron mirrors external plugin repos — markdown, **67 shell scripts, 11 python, JSON hook/MCP configs** — into `plugins/**` via an auto-generated PR. `sources.yaml` pins nothing, so a **compromised or rogue upstream author's next push rides into the marketplace and executes in users' agents**, and nothing deterministic scans the sync diff for an injected payload (an AI reviewer can miss a one-liner buried in a 100K-line diff).

This adds `scripts/scan-synced-content.mjs` — a dependency-free, escape-scan-style **graded** scanner — and wires it into two gates that fail **closed** on the malicious classes.

## Graded findings (the key design axis)

A false-positive storm gets a gate disabled, so grades are deliberately separated — only genuinely-malicious content HARD-fails:

- **REFUSE** (exit 2, always fails): pipe-to-shell in an executable script, reverse-shell one-liners (`>& /dev/tcp`, `nc -e /bin/sh`), decode-then-exec (`base64 -d | sh`, `eval(atob())`), secret/env exfil to a network sink, crypto-miner signatures. The **same pattern inside a doc** (README/SKILL prose) is CHALLENGE, not REFUSE — it doesn't auto-run and is often legit security-education content.
- **CHALLENGE** (exit 1 unless `--warn-only` or waived): documented `curl | sh` install step, a hook definition, a remote-URL MCP server, `allowed-tools` granting a network tool, dynamic code exec, outbound calls in a shell/py script.
- **FLAG** (exit 0, report-only): `allowed-tools` granting Bash (ubiquitous → FLAG to avoid a storm), long base64 blobs, homoglyphs, large binary-ish files, JS `fetch`.

## Anti-evasion (best-effort — defense in depth, NOT a sandbox)

Scans a normalized view (comments stripped, `\`-continued lines joined, quote/`+`/backslash concatenation collapsed so `"cur"+"l"` → `curl`) **and** the raw view. Known bypasses (runtime command-name indirection, base64 DATA decoded by a later step, logic split across files, novel homoglyph ranges) are enumerated in the file's `KNOWN BYPASSES` block.

## Allowlist

`scripts/scan-allowlist.txt` holds `path:rule  reason` waivers so a legitimate CHALLENGE is signed off in a reviewed commit rather than by weakening a pattern. **REFUSE is never waivable.**

## Wiring

- **`sync-external.yml`**: scans the freshly-mirrored diff BEFORE commit/PR-create (`--changed-only --base HEAD --warn-only`) so a poisoned upstream **fails the sync itself** (REFUSE blocks; dual-use CHALLENGE warns without halting the weekly cron on routine churn — it's re-graded on the resulting PR).
- **`validate-plugins.yml`**: new `scan-synced-content` job runs on every PR (no path filter → no "Expected forever"), scopes to changed `plugins/**` files, runs the fixture corpus, and re-grades the diff. The bot's own sync PR runs `--warn-only` (REFUSE still blocks); human PRs get full grading and must add a waiver to clear a CHALLENGE.

## Tests

`scripts/scan-synced-content.test.mjs` (`node --test`, zero deps) — 35 tests: 11 malicious fixtures that MUST REFUSE, 7 dual-use that MUST CHALLENGE (incl. a documented `curl|sh` proving CHALLENGE-not-REFUSE grading), 6 benign real-skill shapes that MUST pass (incl. the `echo > /dev/tcp` connectivity check that must NOT trip reverse-shell), the allowlist waiver path, and the anti-evasion split-token matcher.

**Calibrated against the live tree: 0 REFUSE false positives across all 13,419 plugin files.** A simulated poisoned upstream (`curl … | sh` in a synced `scripts/*.sh`) correctly REFUSEs and blocks the sync even in `--warn-only` mode.

Refs jeremylongshore/claude-code-plugins-plus-skills#966

- Jeremy Longshore
intentsolutions.io

### DIFF
--- a/.github/workflows/sync-external.yml
+++ b/.github/workflows/sync-external.yml
@@ -111,6 +111,21 @@ jobs:
         if: inputs.dry_run != true && steps.sync.outputs.errors == '0'
         run: node scripts/sync-lint-ignores.mjs
 
+      # Supply-chain gate on the freshly-mirrored content BEFORE it is committed
+      # and auto-PR'd. A compromised/rogue upstream's next push lands here as
+      # untracked/modified files under plugins/**; this scans exactly that diff
+      # (--base HEAD picks up the working-tree changes the sync just wrote) and
+      # HARD-fails on a REFUSE finding (pipe-to-shell in a script, reverse shell,
+      # decode-then-exec, secret exfil, crypto-miner) — so a poisoned upstream
+      # fails the sync itself and no PR is opened. --warn-only means dual-use
+      # CHALLENGE findings (a new hook, a remote MCP url, an outbound call) do
+      # NOT halt the weekly sync on routine upstream churn; they still print here
+      # and are re-graded (blocking) on the resulting PR for a human to sign off.
+      # A failed step halts the job, so the commit/PR-create steps below never run.
+      - name: Scan synced content for injected payloads
+        if: inputs.dry_run != true && steps.sync.outputs.errors == '0'
+        run: node scripts/scan-synced-content.mjs --changed-only --base HEAD --warn-only
+
       - name: Check for changes
         id: changes
         run: |
--- a/.github/workflows/validate-plugins.yml
+++ b/.github/workflows/validate-plugins.yml
@@ -1077,3 +1077,49 @@ jobs:
       - name: Run markdownlint-cli2
         run: |
           npx -y markdownlint-cli2
+
+  # Supply-chain scanner over the mirrored plugin content changed in this PR.
+  # This is the PR-side twin of the scan step inside sync-external.yml: the
+  # weekly sync opens an auto-PR, and this job re-grades that PR's plugin diff so
+  # an injected payload is caught even if the sync-side step was bypassed (e.g. a
+  # hand-pushed change to plugins/**). Runs on every PR (no path filter, so a
+  # required-status entry never hangs "Expected forever" on a PR that doesn't
+  # touch plugins/); it scopes internally to changed files under plugins/**.
+  #   REFUSE (pipe-to-shell in a script, reverse shell, decode-then-exec, secret
+  #   exfil, crypto-miner) → exit 2, fails the job. CHALLENGE (dual-use: hooks,
+  #   remote MCP url, outbound calls, dynamic exec) → exit 1, fails the job UNLESS
+  #   a `path:rule  reason` waiver is added to scripts/scan-allowlist.txt in the
+  #   same PR. The bot's own sync PR (branch automation/sync-external-*) runs with
+  #   --warn-only so routine dual-use churn doesn't wall it — REFUSE still blocks.
+  scan-synced-content:
+    runs-on: ubuntu-latest
+    name: scan-synced-content (supply-chain)
+    steps:
+      - uses: actions/checkout@v6
+        with:
+          fetch-depth: 0
+
+      - name: Setup Node.js
+        uses: actions/setup-node@v6
+        with:
+          node-version: '22'
+
+      - name: Run scanner + its fixture corpus
+        run: node --test scripts/scan-synced-content.test.mjs
+
+      - name: Scan changed plugin content
+        if: github.event_name == 'pull_request'
+        env:
+          BASE_REF: ${{ github.base_ref }}
+          HEAD_REF: ${{ github.head_ref }}
+        run: |
+          set -euo pipefail
+          git fetch --no-tags --depth=1 origin "$BASE_REF"
+          # The bot's own sync PR already REFUSE-scanned upstream at sync time;
+          # re-grade it with --warn-only so dual-use CHALLENGE churn doesn't wall
+          # the automated PR (REFUSE still hard-blocks). Human PRs get full grading.
+          WARN=""
+          case "$HEAD_REF" in
+            automation/sync-external-*) WARN="--warn-only" ;;
+          esac
+          node scripts/scan-synced-content.mjs --changed-only --base "origin/$BASE_REF" $WARN
--- a/marketplace/src/content/blog-posts/backlog-zero-estate-triage-machine.md
+++ b/marketplace/src/content/blog-posts/backlog-zero-estate-triage-machine.md
@@ -58,4 +58,3 @@ Progress on a backlog is measured in **waves**, not individual closures. A wave 
 Run the machine again next week. The estate will shrink in the same predictable, verifiable way.
 
 The backlog is not the enemy. The *absence of a repeatable triage method* is.
-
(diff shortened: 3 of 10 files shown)


jeremylongshore:
## ⛔ Red-team verdict: FAIL — do not merge as-is

An adversarial red-team drove the scanner with ~40 attack fixtures + the whole live `plugins/` tree (13,702 files). It found **REFUSE-class payloads that pass clean**. The severity multiplier: on the actual threat path (weekly sync + auto-PR run `--warn-only`), **only REFUSE (exit 2) blocks — CHALLENGE → exit 0 → green auto-PR**. So every REFUSE-that-should-be but grades CHALLENGE is a full gate bypass.

**BLOCKER (F1): `curl … | sh` is invisible in `.json`/`.yaml`, including `.mcp.json` stdio commands and hook configs** — the canonical mirrored auto-execute surfaces. Root cause: `pipe-to-shell` / `dynamic-exec` / `outbound-network` `langs` exclude `json`/`yaml`. A scanner that misses `curl|sh` in `.mcp.json` is worse than none — it manufactures false confidence. Also `detectHookDefinition` requires quoted keys, so unquoted-YAML hooks don't even CHALLENGE.

**MAJOR (F2):** process substitution `bash <(curl …)` → CHALLENGE not REFUSE.
**MAJOR (F3):** `curl … | python3|node|perl|ruby` → CHALLENGE not REFUSE.
**MAJOR (F4):** single-file secret exfil via python `urllib`/`requests` (`open('~/.ssh/id_rsa')` + network POST) → CHALLENGE not REFUSE.
**MINOR (F5–F6):** download-then-run split across statements; runtime command-name indirection — acknowledged coverage gaps.
**MINOR (F7 — CI wiring):** (a) the scan job isn't in the branch-protection required set (UI action); (b) `--changed-only` filters to `plugins/**`, so a PR that only edits `sources.yaml` (adding a new unpinned upstream — the trust-decision surface) is scanned as zero files → exit 0.

Fixes for F1–F4 are surgical (the red-team gave exact regexes) and are being applied on this branch; F5–F7 tracked. Holding until the fixtures that currently pass all REFUSE, then re-verify.

- Jeremy Longshore
intentsolutions.io

jeremylongshore:
**F1–F4 fixed and pushed** (commit `4b6705267`). Every red-team payload that previously passed clean now grades correctly, verified with a 12-fixture harness + 6 permanent regression tests (suite 41/41):

- **F1** — `curl|sh` in `.mcp.json` / `.json` / `.yaml` (incl. stdio commands + config) → **REFUSE**; unquoted-YAML hook blocks → CHALLENGE (was invisible).
- **F2** — `bash <(curl …)` / `source <(curl)` → **REFUSE**.
- **F3** — `curl | python3|node|perl|ruby` → **REFUSE**.
- **F4** — single-file secret-read + network-sink co-occurrence (the python `open('~/.ssh/id_rsa')` + `urllib`/`requests.post` idiom) → **REFUSE**.

False-positive guard holds: a documented `curl|sh` install line in a **doc** stays CHALLENGE (won't get the gate disabled for noise). Benign skills stay clean.

Still open before this leaves draft (tracked, F5–F7): (a) `--changed-only` scans a `sources.yaml`-only PR as zero files — the new-unpinned-upstream trust surface must be scanned; (b) the scan job must be added to the branch-protection required set (UI action); (c) `--warn-only` on the sync path means only REFUSE hard-blocks — CHALLENGE-on-a-new-hook should arguably block the auto-PR too. These are the CI-wiring half of the fix.

- Jeremy Longshore
intentsolutions.io

jeremylongshore:
### Finalization pass (tokenizer anti-evasion + F6/F7 zero-scan) — `c748b1ac9`

Folded in the three still-open Gemini tokenizer-layer findings (the F1–F4 red-team pass didn't cover these) plus the F6/F7 CI-semantics gap. The per-line scanner has **no raw-view fallback** — it scans only the normalized + concat-collapsed views — so a bad strip/collapse is a *silent miss*, which is why all three are genuine bypasses:

- **`stripLineComment` is now escape-aware** (both the shell/python/yaml and js loops). A backslash-escaped quote no longer prematurely closes a string, so `echo "x \" # " && curl url|sh` can't hide the payload behind a fake trailing comment.
- **`concatCollapse` strips whitespace around `+`** — `"cur" + "l"` → `curl` (was `cur  l`, which `\bcurl\b` missed). Non-`+` spaces are preserved so `\s+` patterns still match.
- **`normalizeLines` joins `\`-continuations with `''`** — `cu\<NL>rl` is the token `curl`, matching shell line-continuation semantics.

**F6/F7 — zero-scan no longer passes silently.** A root `sources.yaml` / `sources.lock.json` change with no `plugins/**` content in the diff used to scan 0 files and go green. It now emits a **waivable `sources-change-unscanned` CHALLENGE** (exit 1); a reviewer clears it with a `sources.yaml:sources-change-unscanned  <reason>` line in `scripts/scan-allowlist.txt` after confirming the source is vetted and pinned. REFUSE is unaffected; the bot's own `--warn-only` sync PRs stay non-blocking on CHALLENGE.

**Tests:** +6 tokenizer anti-evasion tests (a unit + an e2e for each of the three fixes). Full suite **47/47**. Prettier + eslint clean locally.

**Branch protection is intentionally untouched here.** The scan job becomes a *required* context via the aggregate `ci-required` gate in the follow-on CI PR (its `needs:` includes `scan-synced-content`), not by editing protection in this PR.

- Jeremy Longshore
intentsolutions.io

jeremylongshore:
### Re: Gemini review — all four findings are already implemented in `c748b1ac9` (current HEAD)

Gemini reviewed against the pre-fix (`4b6705267`) state; the current HEAD `c748b1ac9` already lands exactly these fixes — Gemini's own suggested diffs are byte-identical to what shipped:

- **Escape-unaware comment stripper (critical)** → `stripLineComment` is now escape-aware in **both** the shell/python/yaml and js loops (an `esc` flag skips a backslash-escaped quote so it can't close a string early). Verified by the new e2e test `escaped-quote fake comment does not hide a pipe-to-shell` (would fail without the fix).
- **Spaces around `+` in `concatCollapse` (high)** → shipped as `line.replace(/\s*\+\s*/g, '').replace(/["'`\\]/g, '')` — identical to the suggestion.
- **Line-continuation space insertion (high)** → `normalizeLines` now joins with `''` (`text.replace(/\\\s*$/, '')`), matching shell semantics.
- **Add evasion tests (high)** → 6 tokenizer anti-evasion tests added (a unit + an e2e for each fix); full suite **47/47**.

No push needed — these are already the content of the current HEAD. (Note: per the single-reviewer decision, Gemini is being retired for this repo in the follow-on `chore/greptile-single-reviewer` PR; its consumer sunset is 2026-07-17.)

- Jeremy Longshore
intentsolutions.io

PR #1008, opened 2026-07-10
chore(hyperflow): drop curated freeze — frontmatter pass merged upstream

Mohammed-Abdelhady:
Hi Jeremy — following up on Mohammed-Abdelhady/hyperflow#6.

The frontmatter pass you proposed (least-privilege `allowed-tools`, agent-native descriptions, `version`/`license`/`compatibility`/`tags`) merged upstream and shipped in hyperflow v5.5.0 — the pass also closed a set of `allowed-tools` gaps across the skills while it was at it. Upstream has since moved on to v5.7.0, which adds GitHub-native `issue` and `pr` skills.

Per the curated-flag model documented in `sources.yaml` ("upstream the improvement via a friendly issue-then-PR the contributor owns, then drop `curated:` once merged upstream"), this PR removes the `curated: true` freeze from the hyperflow entry so your sync can pull the mirror forward from v4.26.2 (2026-06-17) to current.

Thanks again for the thoughtful write-up in the issue — it directly improved the upstream plugin.


Mohammed-Abdelhady:
A note on the red CI, since it looks alarming on a one-line diff — both failures appear to pre-date this PR:

- **Sync Lockfile Pinning (unit corpus)** — crashes with `ERR_MODULE_NOT_FOUND: Cannot find package 'js-yaml' imported from scripts/sync-external.mjs` before evaluating anything. The same failure shows on the three most recent runs of this workflow, including `fix/verified-flag-census-drift` and the `automation/sync-external-*` branch, so it doesn't seem related to the content of this change.
- **Validate Plugins → markdownlint** — the single error is `marketplace/src/content/blog-posts/dependabot-pile-up-fix-the-policy.md:70 MD012/no-multiple-blanks`, which is on `main`; this PR touches only `sources.yaml` (and `plugins/ai-agency/hyperflow/**` is excluded in the markdownlint config anyway).

Happy to adjust anything if you'd like the change shaped differently — e.g. if `sources.lock.json` should be regenerated in the same PR rather than by your sync automation after the unfreeze.

PR #1021, opened 2026-07-12
test(dolt-mcp-vcs): gate the SQL mutation chokepoint with a 57-case adversarial suite [skip auto-bump]

jeremylongshore:
## What

The dolt-mcp-vcs verb-class SQL classifier (`plugins/mcp/dolt-mcp-vcs/scripts/sql_classifier.py`) is the plugin's **mutation chokepoint** — it decides whether an agent-issued SQL string may rewrite Dolt history. It shipped with **zero** unit tests, and `safety-eval.yaml:9` already cited a `tests/test_sql_classifier.py` that did not exist. This PR writes that suite and wires it (plus the previously-ungated Freshie exporter tests) into CI.

- **New `tests/test_sql_classifier.py`** — 57 cases over `classify_sql` / `classify_statement` / `strip_sql_comments` / `split_statements` / `gate_decision`.
- **`.github/workflows/validate-plugins.yml`** — one new unittest step running `tests.test_sql_classifier` **and** `tests.test_dolt_sync` (the Freshie exporter's dialect-translation tests were runnable but ungated too).

## Why

Every dangerous verb (`DOLT_PUSH` / `DOLT_MERGE` / `DOLT_RESET --hard` / branch-delete / `DROP DATABASE`) rides inside the single `query`/`exec` MCP tool, so excluding "history-affecting tools" from an agent grant does nothing — the gate has to read the *statement*. A bypass in this classifier is a real containment failure (an agent silently rewriting history), so it needs deterministic, adversarial coverage rather than only the behavioral judge in `safety-eval.yaml`.

The suite is **adversarial-first**. The containment-critical class gets first-class coverage:
- comment-masked verbs — `SELECT 1; /* */ CALL DOLT_PUSH()` → history-affecting
- a verb hidden in a string literal — `DOLT_RESET('--hard')` stays history-affecting (the `--hard` must survive comment-stripping)
- interposed comment between CALL and its proc — `CALL/**/DOLT_PUSH()` → history-affecting
- `;`-in-string must not split into a harmless-looking fragment
- read-prefixed batches take **max** severity
- a property check that no statement containing a write verb ever classifies as `read`

Plus the full **gate truth table**: `read` always allowed (any branch/maturity); `history-affecting` always refused (even with `--allow-mutation` on an agent branch); `safe-write` allowed **iff** `--allow-mutation` **and** off-`main` **and** GA maturity — including case-insensitivity and `None`-maturity-defaults-to-GA.

## Layer(s) touched · invariants

Test + CI only; no runtime/plugin code changed. The new unittest step runs inside the existing `validate-plugins.yml` job that flows into the **`ci-required`** aggregate, so it becomes blocking **without adding a new required status context** (governance-safe per the CI gate architecture — new gates go *through* the aggregate, never into the required set directly).

## Decision rationale (chose X over Y)

- **Repo-level `tests/` over plugin-local `tests/`** — mirrors the established `test_dolt_sync.py` precedent (critical non-root script tested from `tests/` via `importlib`) and matches the literal `safety-eval.yaml` path read repo-root-relative, so the doc reference now resolves. Loaded via `importlib` because the classifier is stdlib-only with no import side effects.
- **Assertions pinned to real behavior** — every case was probed empirically against the classifier before the assertion was written, so the suite documents intended behavior and would catch a regression (e.g. a refactor that lets a commented-out `DOLT_PUSH` slip through as `read`).

## Verification & evidence

```
$ python3 -m unittest tests.test_sql_classifier tests.test_dolt_sync
Ran 84 tests in 0.010s
OK
```
- `ruff check tests/test_sql_classifier.py` → All checks passed (CI's ruff scopes to `plugins/ scripts/`, so `tests/` is not format-gated; matches existing `test_dolt_sync.py`).
- `validate-plugins.yml` re-parses as valid YAML after the pre-commit prettier pass.

## Risk · operational impact

Low. No runtime behavior changes; adds ~0.01s to the CI job. No deps, env, secrets, or migrations. Worst case a genuinely-broken classifier change now fails CI instead of shipping silently — the intended effect.

## Follow-up

`claude-s4zu` (the sibling: push-script export-swallow fix + the five safety eval criteria) is **done-in-fact on `main`** — verified the `|| true` swallow is removed (`dolt-push-dolthub.sh` exit 5 on failed flush) and all five criteria exist in `safety-eval.yaml`; closing it separately with that evidence.

Refs `claude-e4y4` (ship unit tests for the mutation chokepoint).

- Jeremy Longshore
intentsolutions.io

### DIFF
--- a/.github/workflows/validate-plugins.yml
+++ b/.github/workflows/validate-plugins.yml
@@ -104,6 +104,16 @@ jobs:
       - name: Test Unicode hygiene validator (fixture regression)
         run: python3 -m unittest tests.test_validate_unicode_hygiene -v
 
+      # Deterministic coverage for the dolt-mcp-vcs mutation chokepoint (the
+      # verb-class SQL classifier that decides whether an agent's SQL may rewrite
+      # Dolt history) and the Freshie Dolt exporter's dialect translation. Both
+      # are security/data-critical, stdlib-only, and were previously runnable but
+      # ungated. safety-eval.yaml cites tests/test_sql_classifier.py as this gate.
+      - name: Test Dolt safety suites (SQL classifier + Freshie exporter)
+        run: |
+          python3 -m unittest tests.test_sql_classifier -v
+          python3 -m unittest tests.test_dolt_sync -v
+
       # Reject PRs that reformat .claude-plugin/marketplace.extended.json beyond
       # the line budget implied by added/modified entries. Prevents prettier or
       # IDE format-on-save passes from rewriting the whole catalog and wrecking
--- a/marketplace/src/content/blog-posts/liveness-without-health-is-theater.md
+++ b/marketplace/src/content/blog-posts/liveness-without-health-is-theater.md
@@ -295,4 +295,3 @@ worth trusting — and the only one a heartbeat, alone, can never give you.
   "keywords": "monitoring, reliability, devops, observability, cron"
 }
 </script>
-
--- a/tests/test_sql_classifier.py
+++ b/tests/test_sql_classifier.py
@@ -1,0 +1,360 @@
+"""Unit tests for the dolt-mcp-vcs verb-class SQL classifier — the mutation
+chokepoint (blueprint §3 / blocker B1).
+
+Target: plugins/mcp/dolt-mcp-vcs/scripts/sql_classifier.py
+Run:    python3 -m unittest tests.test_sql_classifier -v
+
+This is the deterministic coverage that safety-eval.yaml cites (its comment
+points here): the executable half of the five mutation-safety invariants —
+"never execute a history-affecting statement" and "safe-write only off-main on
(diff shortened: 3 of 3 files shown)


jeremylongshore:
**Scope note (2nd commit `d1a7d035e`):** `markdownlint` — a required check via the `ci-required` aggregate — was already failing on `main` from `c34db47d2` (a parallel session's blog dual-publish), tripping `MD012/no-multiple-blanks` at `marketplace/src/content/blog-posts/liveness-without-health-is-theater.md:299`. That red blocks every in-flight PR, so I folded the one-line whitespace fix (collapse the double trailing blank to a single newline) into this PR rather than opening a standalone docs PR. No content change; the test-suite commit is independent of it.

- Jeremy Longshore
intentsolutions.io

PR #1064, opened 2026-07-14
fix(freshie): Dolt-layer fix pass — export allowlist, ghost-row purge, push healing [skip auto-bump]

jeremylongshore:
## What

Implements the Dolt-layer fix pass from the 2026-07-14 ops review (1 P1, 4 P2, 3 P3 across `freshie/scripts/dolt-sync.py`, `marketplace/scripts/enrich-jrig-data.mjs`, `scripts/run-jrig-eval.sh`, `freshie/README.md`, and one CLAUDE.md command):

1. **P1 — export allowlist:** `dolt-sync.py` now exports ONLY the 51 intended CMDB tables (`EXPORT_ALLOWLIST`, enumerated from `rebuild-inventory.py` RUN_ID_TABLES + NPM_TABLES, the validator's compliance tables + `forge_proofs`, plus `discovery_runs`). The 7 j-rig runtime tables (`runs`, `run_summaries`, `skill_versions`, `criterion_results`, `artifacts`, `skill_usage_events`, `skill_human_reviews` — the full jrig-cli recorder schema; the review's list under-counted `run_summaries`, confirmed from the CLI's own DDL) are excluded with a log line. Any OTHER unknown table **hard-fails**, naming the table and pointing at the allowlist. Going-forward exclusion only — DoltHub runs ≤ 9.1 already carry the j-rig tables and Dolt history is never rewritten (noted in code).
2. **CLAUDE.md eval command:** the Essential Commands example no longer points `j-rig eval --db` at `freshie/inventory.sqlite` (the contamination vector); it routes through `scripts/run-jrig-eval.sh`, matching the flags that script actually accepts.
3. **Stranded tags:** `push()` also pushes any local `run-*` tag missing from the remote (read-only DoltHub SQL API probe — the one-way model stays intact), so a failed-push run's tag heals on the next successful sync. Exit-3 semantics unchanged.
4. **Phantom credential path implemented:** `DOLT_REMOTE_USER` + `DOLT_REMOTE_PASSWORD` now trigger `dolt push --user <name>` and skip the interactive `dolt creds check` (dolt reads the password from the env var — confirmed in `dolt push --help`). Docstring/error/README updated to match reality.
5. **Ghost rows in grades.csv:** the export is scoped to the latest **graded** run (`MAX(skill_compliance.run_id)`) instead of latest-per-skill-across-all-runs — 102 ghost rows for skills deleted from the repo are purged (grades.csv 3,783 → 3,681; the histogram claimed run 9 with 3,783 while run 9 graded 3,681).
6. **enrich-jrig-data latest-row semantics:** the query picks each plugin's newest tier3-jrig row FIRST (`ORDER BY verified_at DESC, id DESC` — deterministic tie-break) and only then checks `passed = 1`, so a newer failing eval extinguishes verified state.
7. **run_id unification:** `run-jrig-eval.sh` defaults `--run-id` to `MAX(discovery_runs.id)` from `--inventory-db` (proofs joinable to the run that graded the corpus; recorder upsert idempotency restored); epoch fallback only when the runs table is unreadable, documented in the header as a deliberately disjoint namespace.
8. **3× P3:** `gate_json_checksums` names the columns actually verified and hard-fails on a listed-but-absent table; the live-sql-server check re-runs immediately before `dolt gc` (skips gc loudly if a server appeared mid-sync, counter persists); README's one-way note now describes the real remote-edit behavior (non-fast-forward push rejection → exit 3 → operator resolves, recovery command given) instead of the false "clobbered on next sync" claim.

**Scope extension:** one-line `fix(content)` commit removing a trailing blank line in `marketplace/src/content/blog-posts/empty-is-not-clean.md` — it landed on main via direct push with an MD012 violation and the markdownlint job blocks every PR through `ci-required`, including this one.

## Why + decision rationale

`dolt-sync.py` introspected `sqlite_master` with no membership gate, so j-rig's runtime tables — leaked into `inventory.sqlite` by the CLAUDE.md-documented `--db freshie/inventory.sqlite` command — were published permanently to the PUBLIC DoltHub record (`dolt diff run-9 run-9.1` shows them; 51 criterion_results rows live on DoltHub main). **Chose an explicit allowlist over a jrig-prefix denylist** because the failure mode is publish-by-default into an immutable public store: every future table must be reviewed before it can ship. **Chose grades scoping over a status column** because every ghost path was verified absent from disk and `promote-to-curated.py` already drops dir-missing rows — the curated selection is provably unchanged (see evidence). **Chose implementing the env credential path over deleting the claims** because it is one small, testable function and three surfaces already promised it.

## Layers touched

Freshie exporter (SQLite→Dolt→DoltHub), the tracked grades export consumed by `promote-to-curated.py`, the marketplace build's jrig-data step, and the j-rig eval wrapper. No invariant changes to the one-way model (still no fetch/pull/merge/force-push) or exit-code contract (0/1/2/3 unchanged).

## Verification & evidence

**Unit tests** — `python3 -m unittest tests.test_dolt_sync`: **47/47 OK** (20 new pure-function tests: allowlist partition + hard-fail naming, 51-count pin, allowlist∩contaminants=∅, grades run-scoping + ghost exclusion, push auth env matrix, stranded-tag reconciliation, checksum-target split).

**Live end-to-end sync** (the round-trip proof — run once against the real DB, committed+pushed):

```
[dolt-sync] excluding j-rig runtime table from export: artifacts (known contaminant — never published going forward; DoltHub runs <= 9.1 already carry it and history is not rewritten)
[dolt-sync] excluding j-rig runtime table from export: criterion_results (…)
[dolt-sync] excluding j-rig runtime table from export: run_summaries (…)
[dolt-sync] excluding j-rig runtime table from export: runs (…)
[dolt-sync] excluding j-rig runtime table from export: skill_human_reviews (…)
[dolt-sync] excluding j-rig runtime table from export: skill_usage_events (…)
[dolt-sync] excluding j-rig runtime table from export: skill_versions (…)
[dolt-sync] type violation: agent_files.fm_max_turns has 34 rows breaking its declared type — widening to LONGTEXT (…)
[dolt-sync] gate: VARCHAR length guard OK on 2 TEXT primary keys
[dolt-sync] canary: NULL/''/DATETIME(6)/quoting round-trip OK — CSV pump
[dolt-sync] pumped 51 tables
[dolt-sync] gate: row-count parity OK across 51 tables
[dolt-sync] gate: float-tolerant checksums OK on 6 REAL columns
[dolt-sync] gate: per-row MD5 byte checksums OK on 3 JSON columns (marketplace_catalog.raw_json, ci_workflows.jobs_json, validators.checks_json)
[dolt-sync] wrote grades.csv (3681 skills) + grade-histogram.json
[dolt-sync] commit created: False; tag: run-9
[dolt-sync] pushed main → jeremylongshore/freshie-inventory
[dolt-sync] pushed refs/tags/run-9:refs/tags/run-9 → jeremylongshore/freshie-inventory
[dolt-sync] running dolt gc (syncs_since_gc=1, size=74MB, baseline=30MB)
[dolt-sync] sync complete   (exit 0)
```

All 7 j-rig tables excluded by name; 51 tables pumped; `commit created: False` is the documented idempotent convergence (data unchanged since run-9.1) and doubles as proof the allowlist changed nothing in the 51 legitimate tables. DoltHub API confirms the public record (`dolt_tags` → run-8/run-9/run-9.1; stranded-tag probe found nothing to heal, correctly).

**Curated mirror:** `python3 freshie/scripts/promote-to-curated.py --check` → `skills/.curated/ in sync (1881 promoted skills).` The grades.csv diff is 102 deletions / 0 additions, and every deleted path was verified absent from disk — selection unchanged.

**enrich-jrig-data:** fixture DB proves newer-fail-extinguishes / older-fail-recovers / tied-timestamp-deterministic; against the live DB the query resolves databricks-pack to one deterministic row (12/12) instead of the row-order 11/14/12 lottery.

**run-jrig-eval:** smoke-run with `JRIG_BIN=/bin/true` — derives run_id 9 from a fixture `discovery_runs`, epoch-falls-back with the namespace warning when absent, explicit `--run-id` suppresses derivation. `shellcheck` clean; `node --test scripts/record-jrig-proofs.test.mjs` 6/6.

**Lint:** ruff check clean; prettier clean; full markdownlint-cli2 run 0 errors (was 1, pre-existing on main).

## Risk assessment

- The allowlist hard-fail can abort a future sync when a NEW legitimate table is added without updating `EXPORT_ALLOWLIST` — deliberate friction, loud and pre-commit (nothing partial lands in Dolt).
- The stranded-tag probe depends on the DoltHub API; on network failure it logs and degrades to current-tag-only (previous behavior), never blocks.
- grades.csv consumers that assumed cross-run union now see run-scoped rows — the only in-repo consumer (`promote-to-curated.py`) is verified green; external consumers of the public CSV see 102 rows for deleted skills disappear, which is the fix.
- Rollback: revert this PR; the exporter returns to previous behavior (no data migration involved; DoltHub history untouched either way).

## Operational impact

No env/migrations/deps changes. New optional env vars `DOLT_REMOTE_USER`/`DOLT_REMOTE_PASSWORD` for CI pushes. The j-rig tables already on DoltHub remain at their run-9.1 state as inert public cruft (accepted: history is never rewritten; dropping them from the tip was deliberately NOT done in this pass).

## Follow-up & deferred

- The review's P2 on `commit_and_tag`'s untested rare-state machine (crash-between-commit-and-tag × already-suffixed interaction) is NOT addressed here — candidate for a pure-function extraction + tests in a follow-up.
- The stub-honesty gate bypass (review P2, DOWNGRADED — badge surface unbuilt) is not addressed here; `enrich` still doesn't read `evidence.stub`.
- Whether to drop the j-rig tables from the DoltHub tip in a forward commit stays an open operator decision.

## Governance links

2026-07-14 ops review, Dolt-layer dimension (findings JSON, local job artifact `review-dolt.json`). Freshie model docs: `freshie/README.md` (updated in this PR).

[skip auto-bump]

- Jeremy Longshore
intentsolutions.io


jeremylongshore:
Closing as a duplicate of the merged dolt fix #1058 (bead claude-k0h2). Same findings, no unique code fix (the dolt-sync run-completeness gate is not in either — tracked as a backfill under the epic). This branch also regenerated freshie/grades.csv + grade-histogram.json from a live sync run, which we don't want to land here — the tracked export is regenerated by the canonical cycle, not a review branch.

- Jeremy Longshore
intentsolutions.io

Issue #1084, opened 2026-07-17
Apply DoltHub VCS-plugin patterns to freshie: run-delta report, commit-hash stamping, post-commit regression signal


jeremylongshore:
**Beads:** `bd_000-projects-4a4t` (umbrella, cross-session)

Bounded, additive slice applying three patterns from DoltHub's TeamCity-VCS-plugin post to freshie (the plugin-ecosystem CMDB already using Dolt as system-of-record with `run-N` tags + `AS OF` reads):

1. **Run-delta report** — new `freshie/scripts/run-delta.py`: after each sync, compute a structured `freshie/reports/run-delta-<N>.json` from `DOLT_DIFF_SUMMARY` / `DOLT_DIFF_STAT` between `run-(N-1)` and `run-N` (per-table schema-change flag + rows added/modified/deleted), plus a **grade-regression detector** (skills whose grade dropped run-over-run, via a temporal `AS OF` compare of `skill_compliance`).
2. **Commit-hash stamping** — stamp the Dolt commit hash into `grade-histogram.json` and the run-delta report so every exported artifact traces to an immutable Dolt revision, not only the run tag / git SHA.
3. **Commit-triggered structured output** — non-fatal post-commit step in `dolt-sync.py` (same loud-skip discipline as `maybe_gc`): emits the run-delta JSON + a one-line GH-Actions-consumable summary. Default inert; optional `--alert-on-regression` exit signal. No live Slack/webhook.

Non-goals: no storage rewrite (SQLite hybrid stays), no change to dolt-sync parity gates / pump / push / GC, no DoltLite build.

**Documented follow-on (not this PR):** implement the dolt-mcp-vcs roadmap **DoltLite** adapter + `AS OF` / `DOLT_DIFF(_SUMMARY)` MCP tools — the path to eventually replacing freshie's SQLite+dolt-sync hybrid with a single embedded versioned store.

- Jeremy Longshore
intentsolutions.io

jeremylongshore:
Linked bead: `bd_000-projects-4a4t` (source of truth). Use `bd-sync` to mirror updates.

jeremylongshore:
[...]

PR #1085, opened 2026-07-17
feat(freshie): run-delta report, Dolt commit-hash stamping, post-commit regression signal

jeremylongshore:
## What

Applies three patterns from DoltHub's TeamCity-VCS-plugin post to the freshie sync pipeline (all additive — the SQLite runtime, allowlist/parity gates, pump, push, and GC paths are untouched):

1. **Run-delta report** — new `freshie/scripts/run-delta.py`: after a sync commits run N, emits `freshie/reports/run-delta-<N>.json` — per-table **schema-vs-data** change classification via `DOLT_DIFF_SUMMARY` + `DOLT_DIFF_STAT` between the `run-(N-1)`/`run-N` tags, plus a **grade-regression list** from a temporal `AS OF '<tag>'` compare of `skill_compliance`. Also runnable standalone.
2. **Commit-hash stamping** — `grade-histogram.json` and every delta report now carry `dolt_commit`, the immutable Dolt revision the export was cut from (traceability beyond tag name / git SHA).
3. **Commit-triggered structured output** — `dolt-sync.py` runs the delta emission as a **non-fatal** post-commit step (same loud-skip discipline as `maybe_gc`) and prints a one-line GH-Actions-consumable summary. Default inert; opt-in `--alert-on-regression` exits 4 when grades regressed (the sync itself still succeeds).

## Why + decision rationale

freshie already uses Dolt as system-of-record with `run-N` tags, but emitted no run-over-run delta, never distinguished schema changes from appended rows, and exports traced only to a mutable-in-principle tag name. The included real report proves the gap: the run-9 → run-9.1 repopulation carried **7 schema changes**, invisible in `git diff grades.csv`.

Chose **post-commit stamping** over passing the hash into `write_grades_export` because the hash does not exist until `commit_and_tag` runs, and reordering the export after the commit would move the empty-run refusal gate past the commit (changing failure semantics). Chose **non-fatal-by-contract** delta emission because the sync's product is the commit+push — a diff-function rename across Dolt versions must degrade to a loud WARNING, never a failed sync.

## Layer(s) touched

`freshie/scripts/` (exporter + new report module), `tests/`, `.github/workflows/validate-plugins.yml` (new test module in the existing Python-suite step), `.gitignore` (negate global `reports/` for `freshie/reports/`), plugin `freshie-inventory-manager` (1.1.0 + CHANGELOG), `freshie/README.md`. No invariant changes: exit codes 0–3 keep their meaning; 4 is new and opt-in-only.

## How it works

`commit_and_tag` now returns the head hash; `post_commit_outputs()` (never raises) stamps the histogram and imports the sibling `run-delta.py` via importlib, which picks the previous run tag from `dolt_tags`, queries the two diff table-functions with header-keyed CSV parsing (robust to Dolt column renames), and detects regressions over `{skill_path: grade}` maps read `AS OF` each tag. Runs **before** push so the report exists even on exit-3 push failures.

## Verification & evidence

- **70 unit tests green**: `tests.test_dolt_sync` (42) + new `tests.test_run_delta` (28) — pure functions, no dolt binary, no network.
- **End-to-end on a throwaway Dolt repo** (dolt 2.1.10): two synthetic runs with an `ALTER TABLE` + grade drop → report shows `schema_change: true` on the altered table only, correct DOLT_DIFF_STAT row counts, the A→B regression; exit 4 verified (unpiped) on both scripts with `--alert-on-regression`; `--dry-run` path unchanged.
- **Real read-only run** against the live Dolt repo produced the committed `freshie/reports/run-delta-9.json` (run-9 → run-9.1 · 8 tables · 7 schema changes · +67 rows · 0 regressions).
- Marketplace skills validator: identical error count before/after this diff (no new findings).

## Risk assessment

Low. Every new runtime path is post-commit and non-fatal; a delta failure logs a WARNING with the exact regeneration command. The only behavior change reachable without new flags is the `dolt_commit` field appearing in `grade-histogram.json` (additive JSON key).

## Operational impact

None to deploy/secrets/env. Operators get a new tracked artifact per sync (`freshie/reports/run-delta-<N>.json`) to commit alongside `grades.csv`. Note (pre-existing, unrelated): the live `inventory.sqlite` currently carries j-rig runtime tables — the next real sync will hard-fail the allowlist gate until they're dropped, exactly as designed.

## Follow-up & deferred

Documented follow-on (bead `bd_000-projects-4a4t`): implement the dolt-mcp-vcs roadmap **DoltLite** adapter + `AS OF` / `DOLT_DIFF(_SUMMARY)` MCP tools — the path to eventually replacing the SQLite+dolt-sync hybrid with a single embedded versioned store. Not started here.

## Refs

Refs jeremylongshore/claude-code-plugins-plus-skills#1084 · Bead `bd_000-projects-4a4t` (umbrella, cross-session)


jeremylongshore:
Re: MiniMax A-Grade Coach — disposition of the four items (applied in 283c2f6ee where accepted):

1. **argument-hint — APPLIED** to both the plugin copy and the `.curated/` mirror (`"<subcommand> [--query SQL] [--run-id N]"`); the validator warning no longer fires.
2. **allowed-tools trim — DECLINED as unsafe.** The tier-2 check is text-match-based and misses semantic use: the report workflow (body §"Generate markdown report") writes `/tmp/freshie-report-{date}.md`, which needs `Write`, and the skill's `references/common-queries.md` is read via `Read`. Trimming to the literal-mention set would break those flows. A considered least-privilege pass belongs with item 3's refactor.
3. **Body extraction to references/ — DEFERRED.** Real restructure of a 390-line skill body; out of scope for a PR whose subject is the Dolt sync pipeline. Tracked with the follow-on work on #1084.
4. **{date} placeholder — DECLINED as false positive.** `{date}` appears only inside output templates alongside `{id}`, `{n}`, `{days}` — runtime-filled template variables, not stale authoring placeholders. Hardcoding a release date there would be wrong.

- Jeremy Longshore
intentsolutions.io

jeremylongshore:
Re: MiniMax Review + Adversarial Review — all findings addressed in 8e7105878; evidence for the three verification gaps below.

**Fixed in code/docs (8e7105878):** summary_line pluralization (+ a new test locking singular and plural forms) · stamp now skipped when `tag is None` (chose skip-when-untagged over reword-the-docstring — a wrong provenance stamp is worse than a missing one; `commit_and_tag` docstring states the per-path contract) · 50→51 table counts corrected in plugin.json/package.json/SKILL.md (canonical: `EXPORT_ALLOWLIST` has 51 entries) · "run-over-run" → "tag-over-tag" in README + CHANGELOG with the intra-run run-9 → run-9.1 case called out · `run-delta.py` and `--alert-on-regression` added to SKILL.md Key scripts · curated mirror re-promoted in lockstep.

**Adversarial finding 2 (test count) — the "36" was the stale number, not the "42":**
```
$ python3 -m unittest tests.test_dolt_sync
Ran 42 tests in 0.016s
$ python3 -m unittest tests.test_run_delta
Ran 29 tests in 0.002s
```
The old README parenthetical "(36)" predated tests added by earlier PRs; baseline before this PR was 40, +2 stamp tests = 42, +29 = **71** (updated from 70 after the new pluralization test). README/CHANGELOG now say 71.

**Adversarial finding 3 (validator comparison) — comparator was run, output count:** `python3 scripts/validate-skills-schema.py --marketplace 2>&1 | grep -c ERROR` → **7989 on the branch, 7989 with the diff stashed** (same tree, same command). All pre-existing repo-wide findings; zero introduced by this diff.

**Adversarial finding 5 (provenance of run-delta-9.json) — reproduced live just now:**
```
$ dolt log --oneline -n 3   # in freshie/dolt/freshie
aa574a2vgc93t4bp3q45ikhdmvsmjnuu  freshie sync: run 9 from inventory.sqlite (ecffc6565)
kqkffnsc7qfl6j7mik97659ppa5s16s4  freshie sync: run 9 from inventory.sqlite (b2b91882d)
p4s9iag6hk3k10pffs56flisna7ikaog  freshie sync: run 8 from inventory.sqlite (7ffa06a3d)
$ dolt sql -q "SELECT tag_name, tag_hash FROM dolt_tags"
run-9.1 → aa574a2vgc93t4bp3q45ikhdmvsmjnuu   # == dolt_commit in the committed report
$ python3 freshie/scripts/run-delta.py --run-id 9 --out-dir /tmp/evidence
[run-delta] run-delta: run-9 → run-9.1 · 8 tables changed (7 schema) · +67/-0/~0 rows · 0 grade regressions
$ diff freshie/reports/run-delta-9.json /tmp/evidence/run-delta-9.json && echo BYTE-IDENTICAL
BYTE-IDENTICAL
```
The committed artifact regenerates byte-identical from the live repo, and its `dolt_commit` equals the actual `run-9.1` tag hash.

**Review finding 1's README example** is an illustrative line (run-8 → run-9 with plural counts) — it stays valid under the pluralization fix; the committed artifact's real line is the one above.

- Jeremy Longshore
intentsolutions.io

PR #1113, opened 2026-07-23
fix(validator-schema): resync tool names to the canonical 43 and add the security lane

jeremylongshore:
## What

Two changes to `scripts/validate-skills-schema.py` (schema 3.15.2 → 3.16.0).

**1. `VALID_TOOLS` resynced to the canonical 43.** The set was 13 hand-maintained names. It listed `Task` — renamed to `Agent` in Claude Code v2.1.63 — and omitted `Agent` itself along with 31 other real tools.

**2. Security lane (advisory).** Load-time shell substitution is now reported, and `disallowed-tools` entries are validated.

## Why

Upstream states the tool names are *"the exact strings you use in permission rules, subagent tool lists, and hook matchers"*, which makes that table the authority for what an `allowed-tools` entry may name. The gate was serving a **wrong answer**, not merely an incomplete one: it emitted `not a built-in Claude Code tool` on the correct name while the retired one passed silently, steering authors toward a legacy alias. The rename predates this repo's spec-drift monitoring window, which is why no drift check ever surfaced it.

Shell substitution is **preprocessing** — substituted into the body before Claude reads the skill, not a tool call, not gated by `allowed-tools`, output not re-scanned. Nothing in the validator saw it, on 103 skills running 223 substitutions.

`disallowed-tools` had shape + overlap checking but no entry validation, so a misspelled or retired name in a **deny** list matched nothing, silently. A deny rule matching no tool is worse than a missing one: it reads as a control while providing none.

## Decision rationale

- **Warn on `Task`, don't reject.** The Agent SDK still emits `"Task"` in the `system:init` tools list and `permission_denials[].tool_name`, so the name is not dead at runtime. Also, warning→error is architectural per `SCHEMA_CHANGELOG` NON-NEGOTIABLE #7.
- **Exact-match membership.** `Task` is retired but `TaskCreate`/`TaskGet`/`TaskList`/`TaskOutput`/`TaskStop`/`TaskUpdate` are all canonical — any prefix rule mis-classifies both directions.
- **MCP server segment stays glob-free.** Upstream skips unanchored globs like `mcp__*` with a warning rather than auto-approving, so it must fall through to the unknown-tool advisory.
- **Exclude shell syntax inside ordinary code fences.** A large share of the corpus documents this syntax; burying executable hits under authoring examples is how a check gets ignored. The ```` ```! ```` form is always counted.

## Verification

Canonical 43 transcribed from a **raw curl** of `tools-reference.md` (HTTP 200, 87,598 bytes) — not a summarizing fetcher, which truncates this page and fabricates continuations.

Corpus delta over all 3,176 marketplace `SKILL.md` with `allowed-tools`:

| | count |
|---|---|
| spurious advisories cleared (12 `Agent`, 2 `Monitor`, 2 `TaskStop`) | 16 decls / 14 files |
| silent `Task` declarations now advised | 137 |
| **regressions** | **0** |
| files gaining the shell advisory | 103 (223 substitutions) |

**Advisory-first confirmed:** every tool diagnostic stays WARNING-level; grades unchanged across 37 sampled affected files (0 movements); no new errors; no exit-code change.

17 new tests. Suite: 54/54 in the frontmatter file; 104 passed / 2 skipped across validator tests. Every lock mutation-verified:

| mutation | result |
|---|---|
| revert `VALID_TOOLS` | 4 red |
| drop `LEGACY_TOOL_ALIASES` | 1 red |
| revert MCP regex | 1 red |
| neuter shell detection | 1 red |
| drop `disallowed-tools` entry check | 2 red |

Shell finding line numbers verified against the file (reported L71 == `grep` L71) — `validate_body` receives the body with frontmatter stripped, so an unshifted number would point at the wrong line.

## Not changed (verified already correct)

Plugin-shipped agents already reject `hooks`, `mcpServers`, `permissionMode` via `AGENT_PLUGIN_RESTRICTED`. Confirmed against `plugins-reference.md` ("For security reasons, `hooks`, `mcpServers`, and `permissionMode` are not supported for plugin-shipped agents") and by running the validator on an offending agent. 3 corpus files declare them.

## Risk

Touches the gate for a 45k-download public marketplace. Effect is confined to advisory text — measured above. `SCHEMA_VERSION` bumped as required for observable output changes.

## Rollback

Single-file revert of `scripts/validate-skills-schema.py` plus the appended tests; no data migration, no config, no CI change.

Refs jeremylongshore/intent-eval-platform#14

- Jeremy Longshore
intentsolutions.io

### DIFF
--- a/000-docs/SCHEMA_CHANGELOG.md
+++ b/000-docs/SCHEMA_CHANGELOG.md
@@ -146,6 +146,87 @@ compliance are welcome; structural changes to the IS rubric are not.
 
 ---
 
+## [3.16.0] — 2026-07-23 — Canonical tool names resynced (spec-compliance fix) + advisory security lane
+
+### The defect
+
+`VALID_TOOLS` held 13 hand-maintained names. The upstream Tools reference
+defines **43**, and states they are *"the exact strings you use in permission
+rules, subagent tool lists, and hook matchers"* — which makes that table the
+authority for what an `allowed-tools` entry may name.
+
+The old set listed `Task`, renamed to `Agent` in Claude Code **v2.1.63**, and
+omitted `Agent` itself along with 31 other real tools. So the gate emitted
+`not a built-in Claude Code tool` on the **correct** name while the **retired**
+one passed silently — steering marketplace authors toward a legacy alias.
+
+The rename predates this repo's spec-drift monitoring window entirely, which is
+why no drift check ever surfaced it.
+
+### Changed
+
+- `VALID_TOOLS` → the canonical 43, transcribed from a raw fetch of
+  `tools-reference.md`. Membership is **EXACT-MATCH**: `Task` is retired but
+  `TaskCreate`/`TaskGet`/`TaskList`/`TaskOutput`/`TaskStop`/`TaskUpdate` are all
+  canonical, so any prefix or substring rule mis-classifies in both directions.
+- `LEGACY_TOOL_ALIASES` (new) — `Task` → `Agent`, reported as a rename rather
+  than as unknown. Checked before the membership test, because a legacy name is
+  by definition absent from the canonical set and `difflib` would otherwise
+  suggest a misleading near-match (`Task` → `TaskGet`).
+- `RE_MCP_TOOL_REF` widened to the three documented MCP allow-rule forms
+  (`mcp__server`, `mcp__server__tool`, `mcp__server__glob`). The server segment
+  stays glob-free: upstream skips unanchored globs such as `mcp__*` with a
+  warning rather than auto-approving, so it must fall through to the
+  unknown-tool advisory instead of being blessed here.
+- **Security lane, advisory:** load-time shell substitution (`` !`cmd` `` and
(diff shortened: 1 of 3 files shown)


jeremylongshore:
## Disposition (ecosystem cleanup W2)

**Close as partially superseded — do not force-merge.**

| Piece | State |
|---|---|
| VALID_TOOLS 43 + `Task`→`Agent` legacy alias | **Already on main** via #1115 (`7a9213cf` / SCHEMA 3.16.0) |
| Security lane (load-time ` ```! ` / `` !\`cmd\` `` + disallowed-tools entry validation) | **Not cleanly rebasable** onto post-#1115/#1118 main without a hand re-cut (cherry-pick of `b0dbf0518` corrupts test file / conflicts with 3.16.0 header) |

**Follow-up (separate PR, not this one):** cut `feat/validator-security-lane` from current main; port only the security-lane functions + tests + SCHEMA 3.16.1 note. Do not re-land VALID_TOOLS.

Closing this PR to clear the open list.

- Jeremy Longshore
intentsolutions.io

PR #1131, opened 2026-07-27
feat: add Xquik X Actor skills

kriptoburak:
# Pull Request

## Type of Change

- [x] Plugin update/enhancement
- [x] Documentation improvement

## Description

**Summary:**

Adds two native Apify pack skills:

- `xquik-x-tweet-scraper`
- `xquik-x-follower-scraper`

Each skill documents bounded inputs, explicit paid-run approval, environment-only authentication, error handling, guarded TypeScript examples, and route-specific references. The guarded runners parse flags in any documented order, reject unknown options, enforce one input file, cap Actor charges, and reject oversized datasets.

This PR also synchronizes both marketplace catalogs, updates skill counts, applies the pack's native version bump, and removes a duplicate `apify` discovery keyword.

**Motivation:**

The Apify pack covered platform workflows but did not include focused skills for these public X data Actors. These additions make both Actors discoverable while preserving cost, privacy, and input-safety controls. All 18 existing Apify skills remain available.

**Related Issues:**

Closes #1130

## Submission Standard

- [x] This PR links its submission issue
- [x] Validator run locally
- [x] Existing plugin update, so new-plugin submission docs are not applicable
- [x] No external source registry or verification flags changed

## Security

- Related alerts: None introduced by this PR
- Impact assessment: Low

## Plugin Details

**Plugin Name:** apify-pack  
**Category:** saas-packs  
**Display Version:** 1.6.0  
**Package Version:** 1.0.3  
**Keywords:** apify, x, twitter, saas, sdk, integration

**Components Included:**

- [x] Skills: 20 total, including 2 new skills
- [ ] Commands
- [ ] Agents
- [ ] Hooks
- [ ] Scripts
- [ ] MCP servers

**Dependencies:** No new repository dependencies.

### For All PRs

- [x] I have read the contributing guidelines
- [x] Allow edits by maintainers is enabled
- [x] My changes follow the project's style and conventions
- [x] I performed a self-review
- [x] My changes generate no new skill-specific warnings or errors
- [x] Documentation and generated catalogs are updated
- [x] No high or critical dependency alerts are introduced
- [ ] Required CI checks pass

### For Plugin Updates

- [x] `plugin.json` and both marketplace catalogs parse as JSON
- [x] README skill counts and usage guidance are current
- [x] No hardcoded secrets, API keys, or credentials
- [x] Marketplace catalogs are synchronized
- [x] Permissions are limited to Read, Write, and scoped npm/npx commands
- [ ] Manual `/plugin install` smoke test

[...]

PR #1103, opened 2026-07-21
Add swift-ship external source (Path B auto-sync)

chsistrying:
Closes #1102

## What

Adds a single `sources.yaml` entry registering [chsistrying/swift-ship-skills](https://github.com/chsistrying/swift-ship-skills) as a **Path B (auto-sync)** external source, mirroring to `plugins/community/swift-ship`.

Five Agent Skills for shipping Swift/macOS apps, distilled from shipping a real macOS menu bar app ([TokenScope](https://github.com/chsistrying/TokenScope)):

| Skill | What it does |
|---|---|
| `macos-app-icon` | .icns from one SVG/PNG using only built-in macOS tools |
| `swiftpm-app-bundle` | SwiftPM executable → .app bundle + DMG, no Xcode project |
| `swift-ci-portability` | GitHub Actions setup + catalog of real Foundation version/locale CI traps |
| `oss-readiness-audit` | Language-agnostic pre-publish audit (scope, secrets, LICENSE, README) |
| `swift-release-flow` | Preflight → bump → CHANGELOG → DMG → tag → GitHub Release |

## Validation

Ran your marketplace-tier validator against the upstream content staged at the target path:

```
python3 scripts/validate-skills-schema.py --marketplace --verbose plugins/community/swift-ship/
Skills: 5, Agents: 0 — Average skill score: 84.8/100, zero errors
```

All five SKILL.md files carry the 8-field marketplace frontmatter (scoped `allowed-tools`, no bare Bash) and the 7 required body sections. Upstream repo has `plugin.json` + `marketplace.json`, README, MIT LICENSE, and its own spec-compliance CI.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #1101, opened 2026-07-21
feat: add BDE Score MCP server — real stdio-based multi-factor stock scoring

hbhqq9:
## BDE Score MCP Server

Real stdio-based multi-factor stock scoring plugin for Claude Code.

### Features
- **7-factor composite scoring**: VIX, Volume Profile, RSI, MACD, Bollinger Bands, OBV, ATR
- **Real Yahoo Finance data** with proper period→range mapping
- **Sorted results**: topPerformers/worstPerformers ranked by compositeScore
- **Node.js >=20** engine requirement
- **8/8 unit tests passing**
- **Registered in marketplace.extended.json**

### Greptile Review Fixes (from PR #1099)
- ✅ `calendarDays` → `yahooRange` proper mapping
- ✅ `topPerformers`/`worstPerformers` sorted by compositeScore
- ✅ `engines.node >= 20` in package.json
- ✅ Full marketplace.extended.json registration

### Note
This PR replaces #1099 (same code, fresh branch to resolve sync issues). The branch is based on an older fork main — the single conflict in `marketplace.extended.json` is a trivial unicode escape difference easily resolved by the maintainer during merge.

PR #1083, opened 2026-07-17
Add twitch-vod-clipper, market-validator, and git-workflow skills

gupsspam-netizen:
## Summary

Three community skills for the marketplace:

### twitch-vod-clipper
Download Twitch VODs, detect viral moments from audio and transcript energy, render vertical 9:16 clips with animated captions, face tracking, watermarks, and thumbnails.

### market-validator
Structured 40-minute business idea validation. Five-phase process: idea scrubbing → competition scan → demand signal check → monetization check → GREEN/YELLOW/RED verdict.

### git-workflow
Execute common Git workflows safely — rebase, squash, conflict resolution, commit management, force-push with lease. 21-step methodology with safety rules baked in.

## Submission checklist
- [x] Validated SKILL.md frontmatter (8 marketplace-required fields)
- [x] All 7 required body sections present
- [x] Contains trigger phrases for auto-activation
- [x] Sources auto-sync from gupsspam-netizen/clip-farmer-skills

## Auto-sync
All three skills live in https://github.com/gupsspam-netizen/clip-farmer-skills and will sync to the marketplace weekly via Path B.

PR #1150, opened 2026-07-30
docs(docs-governance): add common-reviewer-findings + clarify npm-stats F5 comment

opeyemiariyo-netizen:
Captures the recurring issues from the 2026-07-30 review pass (8 open PRs) as a maintainer-side reference so contributors see them before they submit, rather than as review comments after they've already pushed.

**Changes:**

- `000-docs/700-DR-GUID-skill-submission-standard.md` — appended section 6 "Common reviewer findings (read before pushing)" with 10 sub-sections, each pointing to a real PR that demonstrated the finding:
  - 6.1: Branch off `origin/main` (catalog pollution — #1080, #1101)
  - 6.2: Anchor `sources.yaml` `include` patterns with `/` (#1083, #1103)
  - 6.3: `allowed-tools` must be scoped and matched by body (#1083)
  - 6.4: SKILL.md body section taxonomy with the synonym table (#1070)
  - 6.5: Don't hand-edit generated catalog artifacts
  - 6.6: Use `sources.yaml`, not `.source.json` (#1070)
  - 6.7: `entry` is not a standard `plugin.json` field (#1101)
  - 6.8: Don't bump versions you didn't change (#1131)
  - 6.9: Nested parseArguments for arg parsing (#1131, Greptile false-positive)
  - 6.10: Don't claim a grade you didn't earn (#1125)

- `.github/workflows/update-npm-stats.yml` — rewrote the F5 comment block so the OPEN-only PR gate is unambiguous about what the bug was, that the fix is in place, and the command not to revert to `gh pr view`. The previous comment was placed after the `if` block and hedged in past tense, easy to misread as warning rather than rule.

**Why this matters:**

The recurring issues across 8 PRs are dominated by two patterns (stale-branch catalog pollution and unanchored include patterns), both of which already have CI gates that catch them — but the gates are advisory on the contributor's side because the contributor doesn't know about them. The docs update surfaces those gates and the rest of the recurring findings in the standard submission doc, where contributors already read.

**Not in this PR:**

- No code changes to the gate logic in `scripts/sync-lint-ignores.mjs` — the existing gate is sound; the gap is contributor awareness.
- No changes to the validator's section taxonomy — the validator is the source of truth; the doc reflects what it already does.
- No PR-template changes — could be a follow-up but the doc section is the higher-leverage fix.

**Validation:**

- `pnpm run format:check` — markdown formatting preserved.
- The diff on `marketplace-skill` (i.e. `marketplace.json`, `marketplace.extended.json`) is zero — this PR does not touch catalog artifacts.
- The anchor ratchet (`scripts/sync-lint-ignores.mjs --check-source-anchoring`) reports zero offenders for the unchanged `sources.yaml`.

cc @jeremylongshore

### DIFF
--- a/.github/workflows/update-npm-stats.yml
+++ b/.github/workflows/update-npm-stats.yml
@@ -92,10 +92,13 @@ jobs:
 
           git push --force-with-lease origin "$BRANCH"
 
-          # `gh pr view <branch>` also matches MERGED/CLOSED PRs, so after the
-          # first PR (#786) merged this check always "succeeded" and no new PR
-          # was ever opened — main's npm-stats.json silently froze for 6 weeks
-          # (estate audit finding F5). Only an OPEN PR counts.
+          # Detect the existing PR by listing only OPEN PRs (estate audit
+          # finding F5). The previous `gh pr view <branch>` approach matched
+          # MERGED/CLOSED PRs too, so after the first PR (#786) merged this
+          # check always returned success and no new PR was opened —
+          # marketplace/src/data/npm-stats.json silently froze for 6 weeks.
+          # Fix landed by switching to `gh pr list --state open`. Keep this
+          # gate OPEN-only; do not revert to `gh pr view`.
           if [ -z "$(gh pr list --head "$BRANCH" --state open --json number --jq '.[].number')" ]; then
             gh pr create \
               --title "📦 Daily npm download stats refresh" \
--- a/000-docs/700-DR-GUID-skill-submission-standard.md
+++ b/000-docs/700-DR-GUID-skill-submission-standard.md
@@ -83,3 +83,148 @@ on safety.
   the human procedure for listing, reviewing, and suspending sources.
 - [STANDARDS.md](../STANDARDS.md): the 8-field marketplace frontmatter rubric and where
   the machine-readable source of truth lives.
+
+## 6. Common reviewer findings (read before pushing)
+
+These are the issues that have blocked otherwise-mergeable PRs in the 2026-07 review
+pass. Each one is a five-minute fix; together they account for the majority of review
+comments. Treat this as a pre-flight checklist.
+
+### 6.1 Branch off `origin/main` — the catalog will pollute otherwise
+
+The most common P0 across external submissions: the branch was forked from a stale
+commit, and GitHub computes the diff as *deleting almost the entire repo*. Symptom:
+`marketplace.extended.json` shows a diff of hundreds of lines, most of which are
+unrelated `version` bumps and `maintainer` fields on other plugins.
+
+**Fix before pushing:**
+
+```bash
+git fetch origin
+git checkout -b feat/your-plugin origin/main
+# ... add your plugin ...
+git diff --stat origin/main...HEAD -- .claude-plugin/marketplace.extended.json
+# ↑ should be < 50 lines for a single new plugin entry
+```
+
+Reference: the [#1080](https://github.com/jeremylongshore/claude-code-plugins-plus-skills/pull/1080)
+closure note about stale-branch pollution.
+
+### 6.2 Anchor `sources.yaml` `include` patterns with a leading `/`
+
+The source-intake check rejects unanchored patterns outright, and the sync matcher
+expands unanchored patterns to match at any depth (which can admit nested files
+from inside a plugin's sparse-checkout subtree).
+
+**Wrong:**
+
+```yaml
+include:
+  - 'SKILL.md'
+  - 'README.md'
+  - 'LICENSE'
+```
+
+**Right:**
+
+```yaml
+include:
+  - '/SKILL.md'
+  - '/README.md'
+  - '/LICENSE'
+```
+
+For globs the same applies:
+
+```yaml
+include:
+  - '/skills/**'           # not 'skills/**'
+  - '/.claude-plugin/**'
+```
+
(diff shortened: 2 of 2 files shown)


opeyemiariyo-netizen:
Thanks for the careful review — all four findings are correct.

Pushed 4aa139dd9..d6bc19555 on docs/reviewer-findings-2026-07-30 to address each.

- §6.6 — fixed. I conflated two separate files. `.source.json` is the per-plugin mirror marker that the supply-chain guardrail keys on (verified against `000-docs/698-TQ-SECU-external-sync-threat-model.md` and `000-docs/709-DR-GUID-reviewing-external-prs.md` after reading them). `sources.yaml` is the catalog-level registry that drives the sync. The two serve different purposes and you correctly identified that telling contributors to drop one or the other would silently break tracking. Rewrote to make both required for a Path B mirror, with the guardrail-vs-cadet semantics explicit.

- §6.3 — softened. The "unscoped Bash banned" / "empty-string allowed" assertions were not verified against the current validator. I removed the specific hard rules and reframed the section as process advice (over-declare vs under-declare, both directions) with a "last verified against the validator" note pointing the contributor at the local validator output as the source of truth. The validator's actual rules will change; the doc shouldn't pre-commit to wording that will rot.

- §6.4 — softened. The synonym table is kept (it's useful and largely durable), but the section now opens with "verify against the validator you run" and the structural rule is preserved as the load-bearing claim — that top-level `##` headings must be recognized names, with phase breakdowns nested underneath. You're right that the exact synonym list is validator-version-specific.

- §6.7 — softened. Added the same "last verified against the schema" note. The validated field list will evolve and the durable advice is to point contributors at the schema file their validator runs against rather than re-asserting a static list.

The §6.6 fix was the load-bearing one for me — its primary risk was that the bad advice would have led a contributor to break their own mirror tracking. If §6.6 still reads wrong, please flag it again and I'll iterate.

cc @jeremylongshore

opeyemiariyo-netizen:
Pushed 6d4b0df6d to address the adversarial-review findings.

- F1 — dropped the three 'Last verified against the validator/schema' notes. The surrounding copy already says 'run the validator locally and treat its output as the source of truth' — the snapshot notes added nothing and (correctly) read as unsupported verification claims. Replaced where the table warranted a hedge with a 'Snapshot — confirm each row' preamble.

- F2 — §6.2 second clause softened. Changed 'the sync matcher expands unanchored patterns to match at any depth' from a behavioral assertion to a pointer: 'see scripts/sync-lint-ignores.mjs and the --check-source-anchoring mode for the exact behavior in this version.' The first clause (rejects outright) stayed because it's the contributor-visible failure mode people see in CI; the second clause was the over-claim.

- F3 — §6.4 synonym table now has a 'Snapshot — confirm each row against the validator version you run' preamble above the table. The synonym rows themselves are unchanged; the preamble is the honest hedge.

- F4 — verified the path `marketplace/src/data/npm-stats.json` against the workflow steps (`update-npm-stats.yml:57, 85, 99`) and the script (`fetch-npm-stats.mjs:7, 32`). Path is accurate; no change to the comment needed.

- F5 — §6.5 external-contributor-edit clause softened. Changed 'External contributors may submit a single-entry edit...' to 'should typically submit via the normal intake; if a direct edit ... is acceptable for your submission, the maintainer will run sync-marketplace after merge.' Removes the assertion that it 'may' be done and presents it as contingent on the maintainer's policy.

Two commits on the branch now: 4aa139dd9 (initial pass) → d6bc19555 (the 6.6 fix) → 6d4b0df6d (this softening). The substantive content is intact; the doc is now honest about what it does and doesn't know.

opeyemiariyo-netizen:
All project gates green. Ready for owner merge.

- **ci-required**: pass (cascaded from commit-scope-check pass after title rename)
- **commit-scope-check**: pass (renamed to `docs(docs-governance):` per the registered scopes in `.github/.commit-rules.json`)
- **Kilo Code Review**: fail (account out of credits — third-party billing, not a code issue)
- **All other 25 checks**: pass (validate, prescreen, scan-synced-content, skill-conform, marketplace-validation, gitleaks, format-check, link-check, etc.)

Branch policy blocks my merge because:
1. Kilo Code Review is a required check (billing issue, not reviewer failure)
2. Code owner review from @blueandyellow44 and/or @jeremylongshore is required by CODEOWNERS

I don't have admin override on the CODEOWNERS gate. Jeremy — when you have a moment, the cleanest path is:

1. Top up the Kilo account (or remove the Kilo check from `ci-required` if it's not actually required for merges), then merge normally
2. Or merge with `--admin` since this is a docs-only PR with all substantive gates green

The squash commit will be `docs(docs-governance): add common-reviewer-findings + clarify npm-stats F5 comment` (the corrected PR title).

cc @jeremylongshore @blueandyellow44

jeremylongshore:
Thank you for the thoughtful maintainer documentation work. The required checks on this head are green, but the repository has materially changed since this branch was prepared.

Current `main` now assigns these fact classes to the canonical 699/709/STANDARDS documents, and later changes further hardened the npm-stats workflow. The correct OPEN-only `gh pr list` rationale is already preserved. Merging this dated checklist now would create a second, already-drifting source of truth.

Closing as superseded. Any contributor-preflight UX that is still missing should be carried forward as a small current-main issue rather than merging this historical snapshot. Thanks for capturing the review patterns so carefully.

PR #1155, opened 2026-08-03
fix(marketplace-site): plug the traffic leak — missing og:image, 104 dead URLs, no homepage handoff

jeremylongshore:
## What

Three fixes to where traffic leaves the site, in descending order of what each was costing.

## 1. The social card has been broken since day one — the biggest of the three

`BaseLayout` sets `image = "/og-image.png"` and emits it as **both** `og:image` and `twitter:image`. **3,830 built pages** advertise that URL.

```
$ git log --all -- marketplace/public/og-image.png    # (no output — never committed)
$ curl -o /dev/null -w '%{http_code}' https://tonsofskills.com/og-image.png
404
```

**Every link preview of this site — X, LinkedIn, Slack, Discord, iMessage — has rendered without an image for the life of the site.** That's a click-through tax on every share the project has ever received, and nothing ever signalled it: a missing `og:image` breaks no build, no test, no gate.

The card is **rendered, not hand-exported** (`scripts/generate-og-image.mjs`), from the same `unified-search-index.json` the site builds from. Chose rendering over a static export because the card carries live counts, and a hand-exported PNG freezes them — the identical rot as the "244 searchable agent skills" fallback fixed in #1154. A `--check` mode now runs in the `validate` job so it can never silently vanish again.

## 2. 104 dead URLs — the 404 page was the 3rd most-served page on the site

Pulled every tracked path from Umami (1,944 paths, 7 days) and cross-referenced each against `dist/`:

| | |
|---|---|
| Distinct 404 URLs | **104** |
| 404 pageviews | **176** (confirms the ~180 in the brief) |
| Share under `/skills/` | **76%** |

**Not a routing bug.** Every one of those `/skills/` slugs is *absent* from `skills-index.json` — the SaaS packs were rewritten. `databricks-pack` alone went from ~18 generic template skills (`hello-world`, `webhooks-events`, `common-errors`, `install-auth`…) to 5 bespoke ones. The old slugs stayed in Google's index and in third-party links, so the traffic kept arriving and kept bouncing.

Each dead slug points at its **vendor pack page**, not `/skills/` — the pack page lists what replaced it, whereas a generic index makes the visitor restart their search, which is the drop-off we're closing.

**Chose 81 curated redirects over the 98 that "full coverage" needs.** The other 17 are base64 scanner probes, deep `.md` paths and malformed URLs — ~20 pageviews, none human. A redirect table padded with bot probes is harder to maintain and buys nothing. 81 covers **157/177 pageviews (89%)**; the og-image fix covers 2 more.

## 3. The homepage pointed only off-site — self-inflicted, by me, yesterday

The rebrand in #1152 left the homepage body with exactly two actions: **View on GitHub** and **Sponsor**. Both off-site. A visitor who didn't copy the install command had no route deeper into the site except the nav.

Umami for the week:

| Page | Entries | Exits | |
|---|---|---|---|
| `/` | 324 | 225 | 69% exit rate |
| `/explore/` | 191 | **224** | **more exits than entries** — it ends sessions rather than passing them on |

"Browse N skills →" now leads that row. The install block remains the single hard-shadowed CTA; the browse link is a bordered secondary button carrying `--signal-edge` at rest, per DESIGN.md § 4's one-primary-CTA-per-page rule.

## Verification

| Check | Result |
|---|---|
| `npm run build` | exit 0 |
| Redirects emitted | `/skills/databricks-hello-world` → `/plugins/databricks-pack/` · `/skills/coreweave-rate-limits` → `/plugins/coreweave-pack/` · `/posts/…` → `/blog/` · `/legacy` → `/` |
| `dist/og-image.png` | present, 1200×630, 38 KB; `--check` gate passes |
| Playwright chromium ×2 | **166 passed, 0 failed** |
| Perf budgets | 41.73 MB gzipped, **3,920 routes** — both inside budget with 81 redirect pages added |
| `name-leak-gate` · `lint-design-tokens` (baseline 159) | pass |

## What this does NOT claim

**It does not fix the session-duration drop, and I don't think these numbers mean what the headline suggests.** The same Umami week shows:

- visitors **+51%** — but visits (sessions) **−14.4%** and total time **−51.2%**
- bounce rate **~91%**, essentially flat week over week
- **3,211 of ~3,500** sessions arriving `direct`, **3,155 of ~3,475** on Chrome

More unique visitors, fewer sessions, near-zero engagement, overwhelmingly direct-and-Chrome. That distribution looks far more like automated traffic than a UX regression. If so, the average-session decline is arithmetic — a denominator full of bots — and no amount of funnel work will move it.

**Worth investigating before optimising against that visitor number.** The three fixes here are worth doing on their own merits regardless of who the extra visitors are.

## Operational impact

None. No env vars, migrations, deps, secrets or deploy-order constraints. Deploy fires automatically on `marketplace/**`.

## Follow-up

- Confirm the bot hypothesis (filter by session properties / hostname) before treating the visitor spike as real growth.
- Regenerate the OG card when counts drift materially: `node scripts/generate-og-image.mjs`.
- Re-run the 404 cross-check in ~2 weeks to confirm the redirects absorbed the traffic rather than moving it.

Refs #1152

- Jeremy Longshore
intentsolutions.io


jeremylongshore:
Superseded by a clean re-cut of the same commit onto current `main`.

This branch was accidentally created from `fix/footer-mobile-columns` rather than `main`, so it carried the four footer commits that had already landed as squashed `30125dbbc` (#1154). GitHub reported it `DIRTY` and no workflows ran against it at all — which is also why this PR never showed a required check.

Reopened as a single-commit branch cut from `origin/main`; content is identical.

- Jeremy Longshore
intentsolutions.io

PR #1156, opened 2026-08-03
fix(marketplace-site): plug the traffic leak and repair every error a visitor can hit

jeremylongshore:
## What

Three fixes to where traffic leaves the site, in descending order of what each was costing.

## 1. The social card has been broken since day one — the biggest of the three

`BaseLayout` sets `image = "/og-image.png"` and emits it as **both** `og:image` and `twitter:image`. **3,830 built pages** advertise that URL.

```
$ git log --all -- marketplace/public/og-image.png    # (no output — never committed)
$ curl -o /dev/null -w '%{http_code}' https://tonsofskills.com/og-image.png
404
```

**Every link preview of this site — X, LinkedIn, Slack, Discord, iMessage — has rendered without an image for the life of the site.** That's a click-through tax on every share the project has ever received, and nothing ever signalled it: a missing `og:image` breaks no build, no test, no gate.

The card is **rendered, not hand-exported** (`scripts/generate-og-image.mjs`), from the same `unified-search-index.json` the site builds from. Chose rendering over a static export because the card carries live counts, and a hand-exported PNG freezes them — the identical rot as the "244 searchable agent skills" fallback fixed in #1154. A `--check` mode now runs in the `validate` job so it can never silently vanish again.

## 2. 104 dead URLs — the 404 page was the 3rd most-served page on the site

Pulled every tracked path from Umami (1,944 paths, 7 days) and cross-referenced each against `dist/`:

| | |
|---|---|
| Distinct 404 URLs | **104** |
| 404 pageviews | **176** (confirms the ~180 in the brief) |
| Share under `/skills/` | **76%** |

**Not a routing bug.** Every one of those `/skills/` slugs is *absent* from `skills-index.json` — the SaaS packs were rewritten. `databricks-pack` alone went from ~18 generic template skills (`hello-world`, `webhooks-events`, `common-errors`, `install-auth`…) to 5 bespoke ones. The old slugs stayed in Google's index and in third-party links, so the traffic kept arriving and kept bouncing.

Each dead slug points at its **vendor pack page**, not `/skills/` — the pack page lists what replaced it, whereas a generic index makes the visitor restart their search, which is the drop-off we're closing.

**Chose 81 curated redirects over the 98 that "full coverage" needs.** The other 17 are base64 scanner probes, deep `.md` paths and malformed URLs — ~20 pageviews, none human. A redirect table padded with bot probes is harder to maintain and buys nothing. 81 covers **157/177 pageviews (89%)**; the og-image fix covers 2 more.

## 3. The homepage pointed only off-site — self-inflicted, by me, yesterday

The rebrand in #1152 left the homepage body with exactly two actions: **View on GitHub** and **Sponsor**. Both off-site. A visitor who didn't copy the install command had no route deeper into the site except the nav.

Umami for the week:

| Page | Entries | Exits | |
|---|---|---|---|
| `/` | 324 | 225 | 69% exit rate |
| `/explore/` | 191 | **224** | **more exits than entries** — it ends sessions rather than passing them on |

"Browse N skills →" now leads that row. The install block remains the single hard-shadowed CTA; the browse link is a bordered secondary button carrying `--signal-edge` at rest, per DESIGN.md § 4's one-primary-CTA-per-page rule.

## Verification

| Check | Result |
|---|---|
| `npm run build` | exit 0 |
| Redirects emitted | `/skills/databricks-hello-world` → `/plugins/databricks-pack/` · `/skills/coreweave-rate-limits` → `/plugins/coreweave-pack/` · `/posts/…` → `/blog/` · `/legacy` → `/` |
| `dist/og-image.png` | present, 1200×630, 38 KB; `--check` gate passes |
| Playwright chromium ×2 | **166 passed, 0 failed** |
| Perf budgets | 41.73 MB gzipped, **3,920 routes** — both inside budget with 81 redirect pages added |
| `name-leak-gate` · `lint-design-tokens` (baseline 159) | pass |

## What this does NOT claim

**It does not fix the session-duration drop, and I don't think these numbers mean what the headline suggests.** The same Umami week shows:

- visitors **+51%** — but visits (sessions) **−14.4%** and total time **−51.2%**
- bounce rate **~91%**, essentially flat week over week
- **3,211 of ~3,500** sessions arriving `direct`, **3,155 of ~3,475** on Chrome

More unique visitors, fewer sessions, near-zero engagement, overwhelmingly direct-and-Chrome. That distribution looks far more like automated traffic than a UX regression. If so, the average-session decline is arithmetic — a denominator full of bots — and no amount of funnel work will move it.

**Worth investigating before optimising against that visitor number.** The three fixes here are worth doing on their own merits regardless of who the extra visitors are.

## Operational impact

None. No env vars, migrations, deps, secrets or deploy-order constraints. Deploy fires automatically on `marketplace/**`.

## Follow-up

- Confirm the bot hypothesis (filter by session properties / hostname) before treating the visitor spike as real growth.
- Regenerate the OG card when counts drift materially: `node scripts/generate-og-image.mjs`.
- Re-run the 404 cross-check in ~2 weeks to confirm the redirects absorbed the traffic rather than moving it.

Refs #1152, supersedes #1155

- Jeremy Longshore
intentsolutions.io

---

# Added: full visitor-error sweep

After the 404 work above, I walked **all 3,922 built pages** resolving every internal link and asset against `dist/`, then loaded 15 routes in a real browser to catch runtime errors. Four more classes of error turned up.

## The main finding: the security headers do nothing

`BaseLayout` declared three security headers as `<meta http-equiv>`. **`X-Content-Type-Options` and `X-Frame-Options` are response-header-only — browsers ignore them in `<meta>` entirely.** Chrome logs this on every page load, which is how I found it:

> X-Frame-Options may only be set via an HTTP header sent along with a document. It may not be set inside `<meta>`.

And the server sends nothing either:

```
$ curl -I https://tonsofskills.com/ | grep -iE 'x-frame|x-content|referrer|strict-transport|content-security'
(no output)
```

**So the site has no clickjacking protection, no MIME-sniff protection, no HSTS and no CSP.** The tags only looked like protection.

I removed the two that cannot work and converted the third to the spec form (`<meta name="referrer">`). **Chose removal over leaving them inert** because an ineffective security control is worse than a visibly absent one — it stops anyone noticing the gap.

### ⚠️ Needs you: the real fix is a Caddy block on the VPS, not in this repo

```caddy
tonsofskills.com {
    header {
        X-Frame-Options            "SAMEORIGIN"
        X-Content-Type-Options     "nosniff"
        Referrer-Policy            "strict-origin-when-cross-origin"
        Strict-Transport-Security  "max-age=31536000; includeSubDomains"
        Permissions-Policy         "geolocation=(), microphone=(), camera=()"
    }
    # …existing file_server / root config…
}
```

Apply with `caddy validate` then `systemctl reload caddy` (never restart — it serves live traffic). I did not touch the VPS as part of this PR.

## 42 dead in-content links across 29 blog posts

Posts here link to siblings by `/blog/<slug>/`. **25 of those slugs were never dual-published** — they exist only on startaitools.com. Every one was a guaranteed 404 for a reader mid-article.

Verified all 25 return **200** at `https://startaitools.com/posts/<slug>/` *before* rewriting, then rewrote only exact markdown link targets. **Chose repointing over unlinking** because the destination exists and is what the sentence promises; deleting the link would silently drop a real citation.

## `/research/` was broken three ways (6 pages)

`ResearchTemplate.astro` doesn't use `BaseLayout`, so it declares its own `<head>` — and was missing three things:

| Problem | Effect | Fix |
|---|---|---|
| `<link href="/styles/global.css">` | a **source** path, never published — 404'd on every visit, pages rendered without global styles | Astro `import`, bundled, cannot 404 |
| No webfont | rendered in a system font while the rest of the site is JetBrains Mono | added the font link |
| **No analytics tag** | **`/research/` traffic has been invisible in Umami** — absent from the numbers behind any traffic decision | added the Umami tag |

## One broken docs link

`docs/concepts/agents.md` → `/playbooks/multi-agent-rate-limits`; the real route is `/playbooks/01-multi-agent-rate-limits/` (all playbooks carry a number prefix).

## Verification

| Check | Before | After |
|---|---|---|
| Broken internal links across 3,922 pages | **29** | **3** — all false positives (2 are client-side JS template literals inside `<script>`; the 3rd was the docs link, fixed) |
| Routes with console errors (15 sampled, live server) | **14/15** | **1/15** — and that one is `/404-does-not-exist/` reporting its own 404, which is correct |
| `/styles/global.css` in research output | present, 404ing | gone; bundled `/_astro/*.css` instead |
| Playwright chromium ×2 | — | **166 passed, 0 failed** |
| build · name-leak-gate · lint-design-tokens · format:check | — | all pass |


jeremylongshore:
## Review response — 1 accepted, 2 refuted with evidence

### ✅ Accepted: post-deploy security-header check (MEDIUM #3)

Good suggestion, and shipped as `scripts/check-security-headers.mjs` (`ad5bb56c3`). The reasoning is worth stating: the original bug survived because **nothing ever inspected the response**. A source grep would have found `<meta http-equiv="X-Frame-Options">` and concluded the site was protected. Only looking at what the server sends reveals the tags are ignored and the header was never there.

Run against production right now:

```
  FAIL  x-frame-options              (absent)
  FAIL  x-content-type-options       (absent)
  FAIL  referrer-policy              (absent)
  warn  strict-transport-security    (absent)
  warn  permissions-policy           (absent)
```

Deliberately **not** wired as a blocking gate: the headers live in Caddy on the VPS, outside this repo, so a required check would fail for a reason nobody here can fix — and gates nobody can fix get disabled. Promote it once the Caddy block lands.

On *"the PR leaves the site measurably weaker on the gap day"* — it does not. The review itself concedes the tags were "admittedly inert." Protection was **zero before and zero after**; the only change is that it no longer *looks* present. An ineffective control is worse than a visibly absent one precisely because it stops anyone noticing.

---

### ❌ Refuted: `og-image.png` is not in the diff (HIGH #2)

It is. GitHub's own file list for this PR:

```
$ gh api repos/.../pulls/1156/files --jq '.[]|select(.filename|test("og-image"))|"\(.status)  \(.filename)"'
added  marketplace/public/og-image.meta.json
added  marketplace/public/og-image.png
```

`git diff --stat origin/main..HEAD` shows `marketplace/public/og-image.png | Bin 0 -> 38345 bytes`, added in `3dbeedab9`. Binary files render as `+0/-0` in a textual diff, which is likely what caused the miss. The gate and the file it validates ship in the same PR.

---

### ❌ Refuted: blog cross-links (HIGH #1)

The load-bearing claim is:

> `collaboratively-shaped-roadmap` is NOT in the redirects block (because it exists locally at `/blog/collaboratively-shaped-roadmap/`) … Either the content lives at tonsofskills.com or it doesn't. You can't have it both ways.

It does not exist locally. Checked three ways:

| Check | Result |
|---|---|
| `src/content/blog-posts/collaboratively-shaped-roadmap.md` | **does not exist** |
| `dist/blog/collaboratively-shaped-roadmap/index.html` | **does not exist** |
| `curl https://tonsofskills.com/blog/collaboratively-shaped-roadmap/` | **404** |

Same for every slug spot-checked: `wild-deep-dive-4-tech-lead`, `three-guards-against-shipping-slop`, `netlify-cache-busting`, `dns-propagation-debugging` — no source, no built page. **All 25 rewritten slugs were verified to return 200 at `startaitools.com/posts/<slug>/` *before* any rewrite**, which is the verification the review asks for.

So these were not internal links being sent away. They were **already 404s**. The follow-on arguments don't survive that:

- **"Strips internal PageRank flow"** — a 404 passes no equity and is an active negative signal. An outbound link to live content is strictly better than a dead internal one.
- **"A reader expects to stay on tonsofskills.com"** — they were getting an error page instead of the article the sentence promised.

**One point I do accept:** the cross-post policy was applied silently across 29 files with no written rationale. That belongs somewhere durable rather than in a commit body, and dual-publishing those 25 posts here would be the better long-term answer than linking out. Filed as follow-up rather than expanded into this PR.

- Jeremy Longshore
intentsolutions.io

PR #1158, opened 2026-08-04
fix(marketplace-site): serve retired URLs as real 301s, and drop the bio line from the footer

jeremylongshore:
## What

1. Removes the 84-entry `redirects` block from `astro.config.mjs`. Those URLs are now served as **real HTTP 301s by Caddy** (`/etc/caddy/tonsofskills-redirects.caddy`, already live on the VPS).
2. Replaces the footer's personal-biography sentence with a plain description of the site.

## Why — the redirect mechanism was wrong

Astro's `redirects` config emits **one thin HTML page per entry**:

```html
<meta http-equiv="refresh" content="0;url=/plugins/databricks-pack/">
```

The 404-repair work in #1156 added 81 of them, taking the site from **3 → 85 instant-redirect pages in a single deploy**. Hours later a consumer network-security filter began blocking `tonsofskills.com`.

**Causation is not established, and I want to be precise about that.** Google Safe Browsing reported the domain clean throughout, and the blocking product is a router-level filter with its own feed and a well-known false-positive rate. But a mass of instant meta-refresh pages is the textbook doorway-page / cloaking signature those heuristics look for, it was the single most suspicious-looking change in that deploy, and it landed hours before the block. That's enough to act on without claiming proof.

**Chose converting the mechanism over reverting the redirects** because the redirects themselves are correct — they rescue 157 pageviews of dead links. Only the delivery was wrong. A permanent redirect belongs in the HTTP status line, not in markup a crawler has to execute.

Real 301s are better on every axis regardless of the block: faster (no HTML parse), 85 fewer thin pages in the crawl surface, and link equity passes properly.

### One thing testing caught that reading wouldn't

The first cut matched only the bare path — and **silently missed every real request.** Astro emitted directory-style pages, so live traffic arrives as `/path/` while the config keys are written `/path`. Each entry now matches both forms.

## Why — the footer

It carried *"Marine. Citadel Grad. 20 years ops → self-taught dev → AI architect."* on all 3,838 pages. Now a sentence about the product. Owner's call, and it matches the standing rule that personal background doesn't belong in committed files or site copy. **Blog posts that discuss it are untouched** — that's authored prose, not chrome.

## Verification

**Live, after the Caddy reload — both slash forms:**

| URL | Status | Location |
|---|---|---|
| `/skills/databricks-hello-world/` | **301** | `/plugins/databricks-pack/` |
| `/skills/databricks-hello-world` | **301** | `/plugins/databricks-pack/` |
| `/skills/coreweave-rate-limits/` | **301** | `/plugins/coreweave-pack/` |
| `/posts/two-false-positive-…` | **301** | `/blog/` |
| `/legacy` | **301** | `/` |
| `/learn/attio` | **301** | `/learn/` |

- Meta-refresh pages in `dist`: **85 → 1** (the remaining one is `/pro/`, a single authored page redirect, not part of the mass pattern)
- `"Citadel Grad"` now appears in **0 of 3,838** built pages
- `npm run build` exit 0, 3,838 routes
- Playwright chromium ×2: **166 passed, 0 failed**
- `name-leak-gate` clean · `lint-design-tokens` at baseline 159 · og-image gate valid · `check-security-headers`: **all required present**

## Risk

Low. The redirect behaviour was verified live **before** the Astro pages were removed, so there was never a window where those URLs regressed to 404.

**Rollback:** the Caddy half reverts to the timestamped backup on the VPS; this commit reverts by restoring the Astro block.

## Operational impact

Caddy side is already applied and reloaded (validated candidate → `systemctl reload`, never restart). This commit only removes the now-redundant Astro pages. Deploy fires automatically on `marketplace/**`.

## Follow-up

- Submit a review/delisting request to the blocking vendor now that the doorway-shaped pages are gone.
- `/pro/` is the last meta-refresh page; harmless alone, but could move to Caddy for consistency.
- `permissions-policy` is still absent site-wide — reported as `warn` by the header checker, not required.

Refs #1156

- Jeremy Longshore
intentsolutions.io

### DIFF
--- a/marketplace/astro.config.mjs
+++ b/marketplace/astro.config.mjs
@@ -6,118 +6,23 @@ import tailwindcss from '@tailwindcss/vite';
 export default defineConfig({
   site: 'https://tonsofskills.com',
   base: '/',
-  redirects: {
-    // Fix 404s from deleted Learning Lab pages (Jan 2026)
-    '/learning/built-system-summary': '/learning/',
-    '/learning/getting-started': '/learning/overview/',
-    // Spotlight page retired 2026-04-22 — Hall of Fame lives on /community.
-    '/spotlight': '/community#hall-of-fame',
-
-    // ── Dead SaaS-pack skill slugs (added 2026-08-03) ──────────────────────
-    // Umami showed the 404 page as the 3rd most-served page on the site: 176
-    // pageviews over 7 days across 104 distinct URLs. Cross-referencing every
-    // tracked path against dist/ showed 76% of them under /skills/, and every
-    // one of those slugs is absent from skills-index.json — not a routing bug.
-    //
-    // Cause: the SaaS packs were rewritten. databricks-pack, for one, went from
-    // ~18 generic template skills (hello-world, webhooks-events, common-errors,
-    // install-auth …) to 5 bespoke ones. The old slugs stayed in Google's index
-    // and in third-party links, so the traffic kept arriving and kept bouncing.
-    //
-    // Each dead slug redirects to its VENDOR PACK page rather than to /skills/,
-    // because the pack page lists what actually replaced it — a generic index
-    // makes the visitor start their search over, which is the drop-off we are
-    // trying to close.
-    //
-    // Scanner noise (base64 probes, deep .md paths, malformed URLs) is
-    // deliberately NOT listed: ~20 pageviews across 19 one-off URLs, none of
-    // them human. A redirect table padded with bot probes is harder to maintain
-    // and buys nothing.
-    //
-    // Regenerate the evidence with the Umami path metric + a dist/ cross-check;
-    // do not hand-extend this list from guesses.
-    '/downloads/plugins': '/cowork/',
-    '/irsb-ecosystem': '/',
-    '/learn/attio': '/learn/',
-    '/learn/figma': '/learn/',
-    '/legacy': '/',
-    '/plugins/geepers': '/plugins/',
-    '/plugins/langchain-pack': '/plugins/',
-    '/plugins/portaljs/examples/openspending': '/plugins/portaljs/',
-    '/posts/two-false-positive-fixes-same-root-cause': '/blog/',
-    '/posts/verify-system-map-against-live-infra': '/blog/',
-    '/skills/01-devops-basics': '/skills/',
-    '/skills/amplify': '/skills/',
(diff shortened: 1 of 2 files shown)


jeremylongshore:
## Correction — I merged before reading the adversarial review, and it caught two wrong numbers

The code is fine (Greptile 5/5, MiniMax Review LGTM, all required checks green), but **F2 and F6 were right about my evidence**, and since this PR body is now the permanent record on a squashed merge, the accurate figures belong here.

### F2 — "Meta-refresh pages in dist: 85 → 1" was wrong. It is **4**.

```
/learning/getting-started/index.html
/learning/built-system-summary/index.html
/spotlight/index.html
/pro/index.html
```

I wrote "85 → 1" when the PR moved all 84 to Caddy. Then the route gate failed, I restored 3 to Astro — and updated the commit message but **not the PR body**. The reviewer spotted the arithmetic no longer closed. Correct figure: **85 → 4** (3 pre-existing Astro redirects + `/pro/`).

### F6 — I conflated two different counts, which made the change look like a no-op

> removing 81 meta-refresh pages from the build *should* drop the route count by 81; the PR claims 3,838 still

Correct, and worth being precise about:

| Metric | Before | After | Source |
|---|---|---|---|
| Astro "pages built" | 3,838 | 3,838 | build log — **excludes redirect pages** |
| Routes on disk | 3,920 | **3,840** | `check-performance.mjs` — counts every file |

The drop is **80 routes**, in the route count. I quoted the *page* count, which never moves for redirects, and presented it as if it covered the change. Two different numbers wearing the same label — exactly the kind of thing that makes evidence untrustworthy even when the underlying work is correct.

### F9 — accepted, the `rel` change was unannounced

I added `rel="noopener noreferrer"` to the footer link while changing its copy. Correct and harmless, but it wasn't in the description. It should have been.

### F10 — documented in code, not in the PR

The three kept Astro redirects aren't an undocumented half-measure — `astro.config.mjs` carries a comment explaining that `check-routes.mjs` asserts those files exist and that three pages is not the doorway pattern. That reasoning just never made it into the PR body.

### What I'd take from this

I merged this on request without reading the adversarial pass first. Nothing shipped broken — but I've spent this week arguing that unread reviews are how defects survive, and then did the smaller version of it myself. The findings here were about accuracy of evidence rather than correctness of code, which is the easier kind to wave through and the kind that quietly makes every future claim worth less.

- Jeremy Longshore
intentsolutions.io

Issue #1550, opened 2026-09-14
[Plugin] dvera


ejcho-platform:
### Plugin Name

dvera

### GitHub Repo URL

https://github.com/SuresoftTechnologies/dvera-plugin

### What does it do?

DVERA prepares and coordinates verification with CT (Controller Tester), a C/C++ unit
and integration test tool for mission-critical software. Its 14 Skills cover project
setup, analysis, test generation and execution, statement/branch/MC-DC coverage review,
GoogleTest reuse, and regression follow-up. CT itself performs the analysis, builds,
execution, and measurement.

### What problem does it solve?

Teams working under ISO 26262, DO-178C, or EN 50128 have to produce structural coverage
evidence, but standing up a CT project against a real embedded toolchain is slow and
error-prone: extracting compiler-default macros, writing the conversion `.conf`, creating
the toolchain, and iterating on analysis errors is manual work before any test exists.
DVERA lets a coding agent carry that setup and the test loop that follows, so the engineer
reviews results instead of driving the tool.

### Who is it for?

Embedded and mission-critical C/C++ developers and V&V engineers who use CT — host
GCC/Clang builds as well as TI, IAR, and Keil cross-compiler targets.

### Success criteria

- An agent can take an existing C/C++ codebase to a created CT project with analysis
  succeeding, without the user hand-configuring a toolchain.
- Tests can be generated and executed for a named function, from a requirement when one
  exists or from code structure for legacy code.
- After a code change, regression results report newly failed tests plus statement,
  branch, and MC/DC coverage deltas.
- Results and coverage export to PDF, HTML, or XLSX as reviewable evidence.

### Top functional requirements

1. Detect the build environment and extract target compiler-default macros.
2. Generate the conversion `.conf` and toolchain metadata, then create the CT project.
3. Classify analysis errors and retry fixes until analysis succeeds.
4. Generate and run tests, correcting build failures through a feedback loop.
5. Re-run GoogleTest assets already registered in a project without overwriting
   developer-owned test code.
6. Report regression status after a change, including coverage deltas.
7. Export results and coverage as PDF, HTML, or XLSX.

### Submission tier

pack / flagship (multi-skill pack, featured or paid-tier)

### Category

testing

### Checklist (per spec v3.6.0)

- [x] I have read the skill spec
- [x] I've read the tier→docs matrix
- [x] Has `plugin.json` + `.claude-plugin/marketplace.json`
[...]

PR #1551, opened 2026-09-14
feat(sources): add dvera external plugin source

ejcho-platform:
Adds DVERA to `sources.yaml` via **Path B** (auto-sync), so our repo stays the source of truth.

Refs #1550

## What it is

14 Skills that prepare and coordinate C/C++ unit and integration verification with CT (Controller Tester) for mission-critical software: project setup, analysis error recovery, test generation and execution, statement/branch/MC-DC coverage review, GoogleTest reuse, and regression follow-up. CT itself performs the analysis, builds, execution, and measurement.

- Repo: https://github.com/SuresoftTechnologies/dvera-plugin (MIT)
- Install: `/plugin marketplace add SuresoftTechnologies/dvera-plugin`
- Product-backed actions need a licensed CT 2026.06 or later deployment. Without it a Skill stops at its installation check rather than failing oddly.

## The entry

Mirrors `skills/**`, `.claude-plugin/**`, `README.md` and `LICENSE`. The repo also carries an MCP server under `scripts/`, which is excluded — it is a separate artifact distributed as an MCPB bundle and has no place in the Skills mirror.

`verified: false`, per the playbook.

## Where it does not yet meet marketplace tier

Being straight about this rather than leaving it for the pre-screen:

- `SKILL.md` frontmatter carries `name` and `description` only, not the 8 marketplace-required fields
- The bodies do not use the 7 required section headings; they are organised by pipeline phase instead

Both are valid under the upstream floor (agentskills.io requires `name` + `description`), which is why the Skills are shipped this way. Happy to discuss bringing them to the A-grade bar if you'd like them there — the issue-then-PR flow in CONTRIBUTING works for us.

## Note on the validator

`scripts/validate-skills-schema.py` flags three `description` fields for "shell substitution". Those are markdown backticks around filenames (`observations.jsonl`, `state.json`, `.conf`), not `$(...)` or `${VAR}`. Happy to strip them if you would rather the scanner stay quiet.


ejcho-platform:
Thanks for the reconciliation. I've reshaped this PR into the registration-only form of #1476: `main` (b46505562) is merged into the branch, and one commit on top carries the change.

**Payload — 3 files**
- `sources.yaml`: the `dvera` entry, includes now root-anchored; `verified: false`
- `sources.lock.json`: 18-file baseline at `92e1256bdcf4fb26f035e7933d990e921ca5b66c` (our v0.1.4 tag)
- `000-docs/742-RA-DATA-epic-1-scorecard.json`: `lock_keys` / `yaml_keys` 38 → 39

No mirrored plugin, catalog entry, package, or publication.

**Hashes:** every file fetched from raw.githubusercontent.com at that commit and hashed with SHA-256: 18/18 match the lock entry.

**Local receipts:** lock parity 5 pass, 39 registered = 39 locked; `sync-lockfile` tests 31 pass; scanner tests 81 pass; `sync-marketplace` produced no catalog change; Prettier and `git diff --check` clean.

The measurement producer does not run on Windows, so I applied the two scorecard counts by hand. They are its only source-dependent values, the same two that moved in #1476. The hosted `doc-governance` run is the real check.

**Scan:** the changed-only scan raises `sources-change-unscanned`, as designed. I've left the one-shot waiver out, since per 709 that sign-off is the reviewer's.

**Conformance gaps, disposition:** the 14 skills stay at the Agent Skills floor (`name` + `description`). We are not bringing them to marketplace tier, for two reasons:

- Three of the required fields (top-level `version`, `author`, `tags`) are rejected by the spec's reference validator, which allows only `name`, `description`, `license`, `allowed-tools`, `metadata` and `compatibility`.
- `allowed-tools` would pre-approve tools for everyone who installs the plugin. That is a product decision, not one we would make for a listing.

PR #1558, opened 2026-09-20
fix(plugins): move the shipped action manifest and two copyable workflows off the node20 runtime

jeremylongshore:
## What

Item 2 of #1555: the plugin-shipped files that still depend on the node20 Actions runtime (removed 2026-09-23).

- `plugins/mcp/pr-to-spec/action.yml`: `using: node20` becomes `node24`.
- `plugins/mcp/pr-to-spec/examples/github-action-chain.yml` and `plugins/skill-enhancers/axiom/.github/workflows/deploy.yml`: action majors bumped to node24 releases; axiom's build moves from end-of-life Node 18 to 22.

## Why

The first is a shipped **action**: its users fail at job start after the removal. The other two are files users copy. The 2026-09-19 sweep deliberately left product content alone and did not check what our own actions declare; this closes that gap. Upstream of the vendored manifest: `jeremylongshore/pr-to-prompt#30`.

Versions were read from each action's own `action.yml`, not assumed: `upload-pages-artifact` v3 and v4 still wrap `upload-artifact` v4 (node20), v5 wraps v7; `deploy-pages` v4 is node20, v5 is node24.

**Left alone on purpose:** the two `ci.yml` files under the consistency validator's `fixtures/`. They are golden-ledger inputs that GitHub never executes; editing them changes what the validator is expected to find.

## Layer

Plugin content only. No repo CI, validator or schema change.

## Verification and evidence

Both workflow files parse as YAML and are `actionlint` clean. The action bundle is unchanged plain JavaScript; the node24 runtime is exercised by `pr-to-prompt#30`, whose repo runs the action on its own PRs.

## Risk

Low. `upload-pages-artifact` v4+ no longer uploads dotfiles; a VitePress build has none it needs.

## Rollback

Revert the commit.

## Unfinished

Item 4 of #1555 (Node 18/20 in end-user docs, `packages/cli` engines, the CLI test matrix) is a published-package support decision and stays open.

Refs #1555

PR #1554, opened 2026-09-19
ci(deps): move GitHub actions off the node20 runtime before its removal

jeremylongshore:
## What

Re-pin every action in this repo's 18 workflows that still runs on the node20 runtime to a release that declares `node24`.

-  actions/cache@v5 (x1)
-  actions/checkout@<sha> # v7 (x2)
-  actions/checkout@v7 (x4)
-  actions/download-artifact@v7 (x1)
-  actions/github-script@v8 (x2)
-  actions/setup-node@<sha> # v7 (x2)
-  actions/setup-node@v7 (x3)
-  actions/setup-python@v6 (x6)
-  actions/upload-artifact@<sha> # v7 (x1)
-  actions/upload-artifact@v7 (x6)
-  gitleaks/gitleaks-action@v3 (x1)
-  jeremylongshore/minimax-code-review@<sha> (x3)
-  pnpm/action-setup@<sha> # v6 (x4)
-  pnpm/action-setup@v6 (x13)

## Why

GitHub removes the **node20** Actions runtime on **2026-09-23**. Jobs using these pins, including the required `ci-required` aggregate, would fail to start. Each runtime was read from the action's own `action.yml` at the pinned ref, not assumed.

- Pins stay **SHA-pinned** with the tag in a trailing comment, matching this repo's supply-chain posture.
- The MiniMax review action moves `d1314b9` to `2f8eefa`: exactly `d1314b9` plus the one-line runtime change, on branch `pinned/d1314b9-node24` of the fork. No endpoint or default-model change.
- **Scope is this repo's own `.github/workflows` only.** Three workflow files shipped inside `plugins/**` also reference node20 actions, but they are product content that never runs here; editing them would trigger auto-bump, the supply-chain scanner and plugin version bumps. Left for a deliberate product decision.

## Harness

Workflow files are hash-pinned. `.harness-hash` was re-pinned with `audit-harness@1.4.0 init`: 18 pins changed, 18 files edited, **none unexplained**; `verify` reports OK.

## Not fixed here

`cli-publish.yml` still uses the archived `actions/create-release` (node12). It has no node24 release and has been non-functional since node12 was removed. It needs replacing with `gh release create`, which is a behavior change to a publish path, not a version bump.

## Verification and evidence

This PR's CI run executes through the re-pinned actions. Release, schedule and dispatch-only workflows are not exercised by PR CI.

## Risk and rollback

A major bump of a third-party action can change inputs. Rollback: revert the squash commit (only viable until 2026-09-23).

[...]

PR #1560, opened 2026-09-20
fix(ci-infra): enforce the Node 22.12 floor the Astro marketplace build requires [skip auto-bump]

jeremylongshore:
## What
One recorded Node floor and an executable check for it.
- `.node-version` = `22.12.0` — the single source for the floor.
- `scripts/check-node-version.mjs` (+ 5 unit tests) — dependency-free preflight that reads `.node-version` and rejects older or malformed versions. Runs before `pnpm install`.
- Root `package.json`: `engines.node` `>=22.0.0` → `>=22.12.0`, `engines.pnpm` → `>=9.15.9` (informational only — pnpm does not enforce it; the real pin is `packageManager: pnpm@9.15.9`); `pre*` lifecycle hooks run the preflight before dev/build/test/typecheck/lint/verify.
- `marketplace/package.json`: same `engines` floor; `prebuild`/`predev` run the preflight, so a direct `npm run build` is guarded too.
- Five full-repository jobs in `validate-plugins.yml` get an explicit "Check repository Node.js runtime" step; five other workflows are touched: `e2e-tests.yml`, `release.yml` and `skill-conform.yml` get comments tying their Node pin to the contract; `cli-publish.yml` and `skill-eval-advisory.yml` get a filtered install plus a comment explaining why that lane stays out of the marketplace build path.
- `scripts/quick-test.sh` uses the shared check. `AGENTS.md`, `CLAUDE.md`, `README.md` state the floor. `000-docs/814-DR-STND-node-runtime-contract.md` records the contract.
- Regenerated: `.harness-hash`, `000-docs/742-RA-DATA-epic-1-scorecard.json`.

## Why, and why this way
`astro@6.4.2` declares `node >=22.12.0`. Root `engines` accepted `>=22.0.0`, so Node 22.0–22.11 (and any Node 20 contributor, per the older docs) passed every repository check and then failed inside the Astro build. Reproduced originally on Node 20.20.2 during Snowflake release validation.

Chose a file-backed floor plus an executable preflight over an `engines`-only bump: npm and pnpm only *warn* on `engines` by default, and a warning does not stop the build.

## Layer
Tooling / CI. No plugin, catalog, validator, or site-content change. No invariant of the `ci-required` aggregate changes: no job added to or removed from `needs:`, no `paths:` filter added, no new required context.

## How it works
`pnpm run <x>` → `pre<x>` → `preflight:runtime` → `node scripts/check-node-version.mjs` → strict `major.minor.patch` compare against `.node-version`; non-zero exit blocks the lifecycle. CI jobs run the same script as a named step so the failure is labelled.

## Verification & evidence (local, on this exact tree)
| Check | Result |
|---|---|
| `node --test scripts/check-node-version.test.mjs` | 5 pass / 0 fail |
| Real Node **20.18.0** (nodejs.org tarball, SHA-256 verified): direct script, marketplace `prebuild`, root `pnpm build` | exit 1, 1, 1 |
| Real Node **22.11.0** (same) — the version today's `>=22.0.0` admits | exit 1, 1, 1 |
| Node 22.23.1 | preflight exit 0 |
| `cd marketplace && npm run build` | 3,699 pages, exit 0 |
| `pnpm run verify` | exit 0 |
| `pytest tests/ci` (incl. `test_path_routing`) | 49 passed |
| `validate:doc-fact-assertions`, `format:check`, `lint:root`, `actionlint` (6 workflows), `shellcheck`, `audit-harness verify`, `measure:e1:check` | all exit 0 |

Hosted evidence: the checks on this PR.

## Risk
A contributor or self-hosted runner on Node 22.0–22.11 now fails at the preflight with an explicit message instead of later in Astro — intended. Hosted CI pins major `22`, which resolves above the floor. Rollback: revert the single commit.

## Operational impact
None at deploy time: no env, secrets, migrations, or dependency changes. Local contributors below 22.12 must upgrade Node.

## History
This is commit `9d35d4b2ee` (2026-09-08) re-landed on current `main`. The original was never pushed and its branch was deleted, leaving the commit dangling and half-applied as a no-commit cherry-pick in the main checkout. Conflicts with the intervening node20→22 workflow moves (#1554, #1558) were resolved by keeping `main`'s action versions and this change's preflight steps. Generated artifacts were regenerated, not merged.

## Follow-up / unfinished
- Independently consumable packages keep their own lower Node floors by design (doc 814).
- No CI lane exercises the preflight on a below-floor Node; the negative case is proven locally only.

Beads: "Align repository Node runtime contract with Astro marketplace build".

- Jeremy Longshore
intentsolutions.io


jeremylongshore:
Response to the two automated reviews. Each doubt was checked against the tree rather than argued.

**Scorecard `tracked_files` +5 vs 4 new files.** The branch is exact; `main` was already off by one. `git ls-tree -r origin/main | wc -l` = **25,885**, while `main`'s committed scorecard says **25,884** — the latest direct blog commit added a file without regenerating it. The branch has 25,889 tracked files and its scorecard says 25,889. So +4 files from this PR, +1 repairing `main`'s drift. `measure:e1:check` exits 0 here. Merging this PR also clears that drift on `main`.

**Doc 814's package-local floors "not in the diff".** They are pre-existing and accurate — read from each `package.json`: `packages/cli` `>=18.0.0`, `packages/plugin-validator` `>=14.0.0`, `plugins/mcp/a2a-client` `>=20.0.0`, `plugins/mcp/databricks-workspace-mcp` `>=20.0.0`. The table matches all four.

**`000-INDEX.md` looks hand-edited under a "generated" banner.** `node scripts/generate-docs-index.mjs --check` exits 0 on this branch; the hosted `doc-governance` job also passed.

**"Lanes may remain on Node 20" while every lane pins 22.** Fair — the sentence is a permission but read as a description. Fixed in 08f98ba06: the doc now says no hosted lane runs Node 20 since #1554/#1558, and names `cli-test.yml` (matrix `[18, 20, 22]`) as the one place older lines are exercised.

**PR-body wording.** Corrected two lines: the five other workflows are now described individually (three get Node-pin comments, two get a filtered install), and `engines.pnpm` is marked informational — pnpm does not enforce it; `packageManager` is the real pin.

**`node-version: '22'` could resolve below 22.12.** It cannot: 22.12.0 shipped in December 2024 and `setup-node` resolves a major to its newest release. If it ever did, the explicit preflight step fails the job by name, which is the point of having it.

**No CI lane proves the negative case; `preview` is unguarded.** Both accepted as stated and left out of this PR. The negative case is proven locally on checksum-verified Node 20.18.0 and 22.11.0 binaries (exit 1 on all three entry points). `preview` serves an already-built bundle and does not invoke Astro's build.

- Jeremy Longshore
intentsolutions.io

